PR #3, opened 2023-08-04
RSDK-3052 :Add github actions for docker and appimages

bhaney:
Two .yaml files are added to the github workflows

- One to create the docker container with the Viam C++ SDK and other necessary libraries
- One to build the AppImage and upload it to GCP


bazile-clyde (approved):
LGTM! Mostly deferring to James for the Docker stuff.

Makefile
```diff
@@ -50,21 +71,27 @@ run-docker: build
 		--cap-add SYS_ADMIN \
 		-it viam-camera-realsense:$(TAG_VERSION)
 
+# Module
+# Creates appimage cmake build.
+# Builds docker image with viam-cpp-sdk and camera-realsense installed.
 package:
-	cd etc && \
+	rm -rf bin | true && \
+	mkdir -p bin && \
```
Otterverse, line 79: Why recreate this directory here? It's unclear what it's used for. If something gets put here within the appimage recipe, then the recipe should probably be removing/recreating the directory.

Makefile
```diff
@@ -50,21 +71,27 @@ run-docker: build
 		--cap-add SYS_ADMIN \
 		-it viam-camera-realsense:$(TAG_VERSION)
 
+# Module
+# Creates appimage cmake build.
+# Builds docker image with viam-cpp-sdk and camera-realsense installed.
 package:
-	cd etc && \
+	rm -rf bin | true && \
```
Otterverse, line 78: Single pipe | is going to pipe STDOUT to STDIN of the "true" command, not do an "or" like it seems you're expecting. Also, shouldn't be needed as the "f" option on rm forces and ignores missing files/folders anyway. (It always returns "success" with force enabled.)

Makefile
```diff
@@ -50,21 +71,27 @@ run-docker: build
...
+	mkdir -p bin && \
+	cd packaging/appimages && \
+	rm -rf deploy | true && \
```
Otterverse, line 81: Same as before, single pipe doesn't do what you think, but also isn't needed.

Makefile
```diff
@@ -50,21 +71,27 @@ run-docker: build
...
+	mkdir -p deploy && \
 	appimage-builder --recipe viam-camera-realsense-aarch64.yml
-
+	cp ./packaging/appimages/viam-camera-realsense-latest-aarch64.AppImage  ./packaging/appimages/deploy/
 
 # Copies binary and AppImage from container to host.
 copy-bin:
```
Otterverse, line 87: Is there a reason to avoid using canon here? This entire target wouldn't be needed as the source folder would be mounted from the host in the first place.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,43 @@
...
+jobs:
+  docker:
+    name: Build Docker Image
+    strategy:
+      matrix:
+        include:
+          - arch: [buildjet-2vcpu-ubuntu-2204-arm]
+            image: ghcr.io/viamrobotics/viam-camera-realsense:arm64
+            platform: linux/arm64
+            make_target: 'docker-cache-arm64-ci'
```
Otterverse, line 26: There's no reason to use a matrix when you only have one target.

.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
...
+    branches: [ 'main' ]
+    paths-ignore:
+      - 'README.md'
+
+# To test workflow updates you need to work in a branch directly on viamrobotics/viam-camera-realsense
+# and tag your working branch instead of @main in any viamrobotics/viam-camera-realsense "uses" below.
+# Don't forget to tag back to @main before merge.
+
+jobs:
+  docker:
```
Otterverse, line 18: What's the point in running this on every push? It should only rebuild if the dockerfile changes I think.

.github/workflows/appimage.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  appimage:
+    name: AppImage Build Intel RealSense
+    strategy:
+      matrix:
+        include:
+          - arch: [arm64, qemu-host]
+            image: ghcr.io/viamrobotics/viam-camera-realsense:arm64-cache
+            platform: linux/arm64
```
Otterverse, line 26: No purpose in "matrix" if there's only one entry.

.github/workflows/appimage.yml
```diff
@@ -0,0 +1,74 @@
...
+    - name: Check out PR branch code
+      if: inputs.release_type == 'pr'
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        submodules: recursive
+
+    - name: apt update
+      run: |
+        sudo apt update
```
Otterverse, line 49: What's the purpose of this? No apt actions are taking place after this.

.github/workflows/appimage.yml
```diff
@@ -0,0 +1,74 @@
...
+      GCP_CREDENTIALS:
+        required: true
+
+jobs:
+  appimage:
+    name: AppImage Build Intel RealSense
+    strategy:
+      matrix:
+        include:
+          - arch: [arm64, qemu-host]
```
Otterverse, line 24: Why not use buildjet here? They're likely faster than our qemu emulated ones (which are going away at some point anyway.)

Otterverse (changes requested):
Got quite a few issues here (sorry!), but all mostly small. Most are noted inline, but additionally, there are some issues not directly modified by this PR with the Dockerfile and the Makefile that concern me.

1. Why not use canon instead of custom docker run/copy commands? It'd make it more consistent with RDK and other repos that way, and simplify some target commands.
2. There's a docker run target in the makefile using CAP_SYSADMIN which effectively breaks docker's security, and isn't good practice. If you must run an appimage for testing inside a docker container, use APPIMAGE_EXTRACT_AND_RUN=1 as documented here: https://docs.appimage.org/user-guide/troubleshooting/fuse.html#extract-and-run-type-2-appimages
3. (minor) The dockerfile runs apt-get update multiple times, which is a bit wasteful.
4. (minor) The dockerfile also downloads appimage-builder from source, rather than just pinning a version with pip.
5. (question) Why does the dockerfile install things to $HOME/opt/src? I'd expect /opt/src or $HOME/src, but $HOME/opt/src feels like some conflation of the two.

bhaney:
> Got quite a few issues here (sorry!), but all mostly small. Most are noted inline, but additionally, there are some issues not directly modified by this PR with the Dockerfile and the Makefile that concern me.

Mostly cargo-culted, and Frakensteined copy-pasted from previous Makefiles and Dockerfiles I have previously used until things compiled. Nice to learn more about how things actually work!

> 
>     1. Why not use canon instead of custom docker run/copy commands? It'd make it more consistent with RDK and other repos that way, and simplify some target commands.

canon for me didn't have the Viam C++ SDK, and also I was getting weird protobuf version errors when I installed the C++ SDK in canon (two different versions of protobuf shared libraries). Just decided to start with the C++ SDK dockerfile and build on top of that/add things until everything compiled. 

>     2. There's a docker run target in the makefile using CAP_SYSADMIN which effectively breaks docker's security, and isn't good practice. If you must run an appimage for testing inside a docker container, use APPIMAGE_EXTRACT_AND_RUN=1 as documented here: https://docs.appimage.org/user-guide/troubleshooting/fuse.html#extract-and-run-type-2-appimages

run-docker isn't used in github actions and was just for my own personal use for debugging compilation errors while building the module. I'll remove the target from the makefile.

>     3. (minor) The dockerfile runs apt-get update multiple times, which is a bit wasteful.

Sometimes you just want to make sure! I'll delete the extra ones and confirm everything still compiles all right. 

>     4. (minor) The dockerfile also downloads appimage-builder from source, rather than just pinning a version with pip.

Will change, copy/pasted from another Dockerfile 
> 
>     5. (question) Why does the dockerfile install things to $HOME/opt/src? I'd expect /opt/src or $HOME/src, but $HOME/opt/src feels like some conflation of the two.

Will change, copy/pasted from another Dockerfile . Though Docker here is just used to compile the binary and create the AppImage, and nothing more, so it seems like what the directory is called doesn't really matter.

Makefile
```diff
@@ -33,14 +33,35 @@ clean-all: clean
...
+
+docker-cache-build: docker-cache-arm64
+
+docker-cache-arm64: MAIN_TAG = ghcr.io/viamrobotics/viam-camera-realsense
+docker-cache-arm64: BUILD_TAG = arm64
+docker-cache-arm64:
+	$(BUILD_CMD)
+
+docker-cache-upload:
+	docker push 'ghcr.io/viamrobotics/viam-camera-realsense:arm64-cache'
```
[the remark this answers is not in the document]
bhaney: copy-paste from RDK workflow, it seems like in that repo there is only a canon-cache? I'm not sure of the difference between canon and  canon-cache. When using canon it only downloads canon-cache. I'll get rid of the "cache" string everywhere

Makefile
bhaney: copy/paste from another Makefile, will remove the second expression

Makefile
bhaney: removed the latter expression

Makefile
bhaney: The canon RDK images didn't have the libraries I needed, so I just used classic Docker to build the app image and copy it out of the container

.github/workflows/appimage.yml
bhaney: will switch to buildjet

.github/workflows/appimage.yml
bhaney: removed

.github/workflows/main.yml
bhaney: removed from this file, and added a on: push: method to the docker.yml file

Otterverse:
> canon for me didn't have the Viam C++ SDK, and also I was getting weird protobuf version errors when I installed the C++ SDK in canon (two different versions of protobuf shared libraries). Just decided to start with the C++ SDK dockerfile and build on top of that/add things until everything compiled.

Sorry, I wasn't clear there. I meant the "canon" CLI utility, not the pre-built images. I totally understand that you need to create your own docker image for this. I meant more the makefile "run-docker" target and stuff. It could just be "canon" if you set up a .canon.yaml here, and then it'd be trivial to drop into the container. Sounds like you're removing that anyway though, so irrelevant now I think.

Makefile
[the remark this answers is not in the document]
Otterverse: Canon itself is built in the SysOps repo. The "-cache" version basically adds about 700MB of pre-downloaded go modules in ~/go/mod/ so that you (and the CI) don't have to download them all every time a container is spun up. So for RDK, there's never really a point for the non-cache use of canon. (But because it's built in two parts, updating the cache layer is far faster than rebuilding the whole image every time.)

.github/workflows/docker.yml
bhaney: changed away from matrix

.github/workflows/appimage.yml
bhaney: done

Makefile
Otterverse: As mentioned in the other comment, I meant use the "canon" cli utility to run your custom image. Then it handles all the magic of setup/teardown for you when you run locally, and stuff built in-place is already available outside the container with no need for extra copy commands and stuff.

Makefile
```diff
@@ -33,38 +33,57 @@ clean-all: clean
...
+docker-arm64-ci:
+	$(BUILD_CMD)
+
 .PHONY: build
 build:
 	docker build -t viam-camera-realsense:$(TAG_VERSION) \
 		--memory=16g \
-		--build-arg TAG=$(TAG_VERSION) \
+	    --build-arg TAG=$(TAG_VERSION) \
 		-f ./etc/Dockerfile.debian.bookworm ./
```
Otterverse, line 65: This target ("build") should go away entirely, right? It's replaced by "docker-arm64" if I'm not mistaken.

etc/Dockerfile.debian.bookworm
Otterverse, line 1: So the last steps in this file should be removed (I can't comment on those lines directly as they're not modified yet.) The viam-camera-realsense source code shouldn't live in the image, nor should it build the actual binary here when creating the image.

Makefile
Otterverse: This "copy-bin" target should go away entirely, as the docker image should no longer contain the actual source code for this repo (which will become out of date quickly.) Instead, you should add a .canon.yaml here configured to use the image you are building. Then all of what was in "copy-bin" simply becomes `canon -arch arm64 make package` and will use the exact source code the user is working with, rather than a potentially obsolete copy from when the image was built.

Otterverse: For the .canon.yaml you should only need a couple lines.

```
viam-camera-realsense:
    arch: arm64
    image: ghcr.io/viamrobotics/viam-camera-realsense:arm64
    persistent: true
```

Makefile
```diff
@@ -33,38 +33,57 @@ clean-all: clean
...
-run-docker: build
-	docker run \
-		--device /dev/fuse \
-		--cap-add SYS_ADMIN \
-		-it viam-camera-realsense:$(TAG_VERSION)
-
+# Module
+# Creates appimage cmake build.
+# Builds docker image with viam-cpp-sdk and camera-realsense installed.
 package:
```
Otterverse, line 69: This should depend on the "camera-module" target, otherwise it's going to package nothing.

etc/Dockerfile.debian.bookworm
Otterverse: Also, near the top of the file, you set "ENV HOME /root" but the problem is that this persists to RUNTIME as well, not just buildtime. That means anything that tries to use another user (such as the canon utility) will fail, as it's home directory will be set to "/root" If you need a variable, us another variable name here. Or just hardcode "/root" in the few places you use it.

Makefile
Otterverse: Caveat... this ALMOST works. You have forced the ENV var "HOME" to be set in the dockerfile though, so canon is broken since the user it creates still has it's home directory forced to root (where it doesn't have permissions.)

Otterverse (changes requested):
A another chunk of suggestions! Sorry I didn't catch some of these on the first pass. On the plus side, I tested building from this branch with canon and it _almost_ "just works" (there's one issue in our dockerfile that breaks things.)

.github/workflows/appimage.yml
```diff
@@ -1,72 +1,40 @@
...
   appimage:
     name: AppImage Build Intel RealSense
-    strategy:
-      matrix:
-        include:
-          - arch: [arm64, qemu-host]
-            image: ghcr.io/viamrobotics/viam-camera-realsense:arm64-cache
-            platform: linux/arm64
-    runs-on: ${{ matrix.arch }}
+    runs-on: [buildjet-2vcpu-ubuntu-2204-arm]
```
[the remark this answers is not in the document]
bhaney: done -- switched to 8. Some compilation and `make install`s should be able to take advantage of more cpus

.github/workflows/appimage.yml
```diff
@@ -1,72 +1,40 @@
...
 
-    - name: apt update
-      run: |
-        sudo apt update
+    - name: Build
+      run: make package
 
     - name: Authorize GCP Upload
-      uses: google-github-actions/auth@v1
+      uses: google-github-actions/auth@v0.4.3
```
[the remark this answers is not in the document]
bhaney: done- mindless copy/paste

.github/workflows/appimage.yml
```diff
@@ -1,72 +1,40 @@
...
-      if: inputs.release_type == 'latest'
-      run: |
-        make package
-
-    - name: Deploy Files
+    - name: Upload Files
       uses: google-github-actions/upload-cloud-storage@v0.10.4
       with:
-        headers: "cache-control: no-cache"
+        headers: "Cache-Control: no-cache"
```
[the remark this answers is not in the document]
bhaney: done

.github/workflows/docker.yml
```diff
@@ -0,0 +1,31 @@
...
+    paths: ['etc/**']
+
+jobs:
+  docker:
+    name: Build Docker Image
+    runs-on: [buildjet-2vcpu-ubuntu-2204-arm]
+    container:
+      image: ghcr.io/viamrobotics/viam-camera-realsense:arm64
+      options: --platform linux/arm64
+            make_target: 'docker-arm64-ci'
```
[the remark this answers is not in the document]
bhaney: done

Makefile
bhaney: done

etc/Dockerfile.debian.bookworm
bhaney: removed the HOME ENV variable, and tested it out with canon - it works now!

Makefile
bhaney: removed the build target

Makefile
bhaney: nicely works with canon now

etc/Dockerfile.debian.bookworm
```diff
@@ -64,17 +58,18 @@ RUN cd ${HOME}/opt/src && \
     rm -rf build
 
 # install viam c++ SDK from source
-RUN cd ${HOME}/opt/src && \
+RUN cd /root/opt/src && \
```
Otterverse, line 61: [nit/optional] This run command could end with `&& rm -rf /root/opt/src/librealsense` to make the image smaller. Each dockerfile command creates a layer, but if you pull in files, AND remove them in the same RUN command, then they don't persist in the layer. Only what was built and installed would remain, thus making the overall image smaller in size.

etc/Dockerfile.debian.bookworm
```diff
@@ -64,17 +58,18 @@ RUN cd ${HOME}/opt/src && \
...
+    git reset --hard e8bddd0dba37c13bfd355b44852a6a5cdbf7baa0 && \
     mkdir build && \
     cd build && \
     cmake  -DCMAKE_BUILD_TYPE=RelWithDebInfo -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON -DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON -DCMAKE_INSTALL_PREFIX=/usr/local .. -G Ninja  && \
     ninja install -j1
 
 # install librealsense from source
-RUN cd ${HOME}/opt/src && \
-    git clone https://github.com/IntelRealSense/librealsense.git && \
+RUN cd /root/opt/src && \
```
Otterverse, line 71: [opt] This too could end in `&& rm -rf ...`

Otterverse (approved):
LGTM! Nice work getting that docker image whipped into shape, and working with canon! Left a couple of (totally optional) notes about how you can reduce the image size, but up to you if you want to bother or not.

PR #6, opened 2023-08-11
upload modules on release

abe-winter:
## what
- upload to module registry on every github release
- (external) updated [package settings](https://github.com/orgs/viamrobotics/packages/container/viam-camera-realsense/settings) to allow writes from this repo
- pinned appimage-builder to most recent commit to fix https://github.com/AppImageCrafters/appimage-builder/issues/272
- increased size of buildjet runner (it's been OOMing)
## notes
- Use version `0.0.0-module-ci.17` if you want to test the built module on a device
- this won't do continuous deployment on main branch (it needs a release). we can set that up as well
- the CI auth will expire in less than 24 hours (known issue for now). there are instructions for setting this [here](https://github.com/viamrobotics/upload-module#setting-cli-config-secret), or ping me

### DIFF
--- a/.github/workflows/appimage-tag.yml
+++ b/.github/workflows/appimage-tag.yml
@@ -26,6 +26,18 @@ jobs:
     - name: Build
       run: make appimage TAG_VERSION="$GITHUB_REF_NAME"
 
+    - name: bundle module
+      run: |
+        cp packaging/appimages/deploy/viam-camera-realsense-*-aarch64.AppImage viam-camera-realsense-aarch64.AppImage
+        tar cf module.tar.gz viam-camera-realsense-aarch64.AppImage
+
+    - uses: viamrobotics/upload-module@main
+      with:
+        namespace: viam
+        platform: linux/arm64
+        version: ${{ github.ref_name }}
+        cli-config-secret: ${{ secrets.cli_config }}
+
     - name: Authorize GCP Upload
       uses: google-github-actions/auth@v1
       with:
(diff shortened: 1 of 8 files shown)


.github/workflows/appimage-tag.yml
```diff
@@ -26,6 +26,18 @@ jobs:
     - name: Build
       run: make appimage TAG_VERSION="$GITHUB_REF_NAME"
 
+    - name: bundle module
+      run: |
+        cp packaging/appimages/deploy/viam-camera-realsense-*-aarch64.AppImage viam-camera-realsense-aarch64.AppImage
```
bhaney, line 31: Potentially, in the future, there might be more than one AppImage that ends up in the deploy directory. What you can do is specifically copy the tagged one by doing 

`cp packaging/appimages/deploy/viam-camera-realsense-"$GITHUB_REF_NAME"-aarch64.AppImage viam-camera-realsense-aarch64.AppImage`

bhaney (commented):
I think there is just one fix to do, and then we can merge this. Thanks for setting this up!

.github/workflows/appimage-tag.yml
abe-winter: fixed, thanks for catch

bhaney (approved):
LGTM

abe-winter:
thanks -- I'll merge after we've done a successful test on hardware

meta.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "viam:realsense",
+  "visibility": "private",
```
bazile-clyde, line 3: Are all modules private right now?

bazile-clyde (approved):
One question but LGTM!

meta.json
bazile-clyde: Yep!

PR #8, opened 2023-10-18
RSDK-5239, RSDK-5388: use CMake and CTest

bhaney:
Tested out on orin nano and raspberry pi: still works!

This PR does NOT add any real tests, but simply changes the compilation structure for the module. The meat of tests will come in the next PR.

This PR includes
- changes the directory structure of the intel realsense module, adding a src and test directory
- creates a header file for the intel realsense 
- introduces the use of CMake for compilation
- introduces a github action for testing using CTest
- changes the dockerfile slightly to install GTest, as well as create a new Dockerfile for cuda (future work)


tests/camera_realsense_test.cpp
bhaney, line 1: This is just a small filler test to make sure the whole pipeline works. Real tests will be added in the next PR

PR #9, opened 2023-10-18
 RSDK-5239, RSDK-5388: use CMake and CTest

bhaney:
Tested out on orin nano and raspberry pi: still works!

This PR does NOT add any real tests, but simply changes the compilation structure for the module. The meat of tests will come in the next PR.

This PR includes:

- changes the directory structure of the intel realsense module, adding a src and test directory
- creates a header file for the intel realsense
- introduces the use of CMake for compilation
- introduces a github action for testing using CTest
- changes the dockerfile slightly to install GTest, as well as create a new Dockerfile for cuda (future work)


tests/camera_realsense_test.cpp
```diff
@@ -0,0 +1,17 @@
+#include "camera_realsense.hpp"
+#include <gtest/gtest.h>
+#include <viam/sdk/components/camera/camera.hpp>
+#include <viam/sdk/components/component.hpp>
+#include <viam/sdk/components/camera/server.hpp>
+
+namespace vsdk = ::viam::sdk;
+
+TEST(ResourceTest, Type) {
```
bhaney, line 9: Just a small, silly test to make sure the CI is working properly. The actual tests will be in the next PR

tests/camera_realsense_test.cpp
```diff
@@ -0,0 +1,17 @@
...
+    vsdk::ResourceConfig testConf = vsdk::ResourceConfig("camera");
+    EXPECT_EQ(testConf.type(), "camera");
+}
+
+int main(int argc, char **argv) {
```
acmorrow, line 14: It looks like you link `GTest::main` above, in which case you probably don't need to declare your own mainline. Or, if you really do want your own main, you can do without linking `GTest::main`.

tests/camera_realsense_test.cpp
acmorrow: I see a potential problem on the horizon. Currently, all the realsense stuff is compiled into an executable. If you want this test to be able to use it, it'll need to be in some sort of library.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# CMakeLists.txt for tests
+
+find_package(GTest REQUIRED)
+find_package(viam-cpp-sdk REQUIRED)
```
acmorrow, line 4: You already did `find_package(viam-cpp-sdk REQUIRED)` in the parent CMakeLists so you probably don't need to say it here too.

src/camera_realsense.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include <viam/sdk/rpc/server.hpp>
+
+constexpr char kResourceType[] = "CameraRealSense";
+constexpr char kAPINamespace[] = "viam";
+constexpr char kAPIType[] = "camera";
+constexpr char kAPISubtype[] = "realsense";
+
+namespace realsense {
+
+namespace vsdk = ::viam::sdk;
```
acmorrow, line 27: If you follow my above suggestion, then I think you can remove this and just say `sdk::`

CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+# construct the executable
+add_executable(viam-camera-realsense src/camera_realsense.cpp ${THIRD_PARTY_SOURCES})
+
+enable_testing()
+add_subdirectory(tests)
+
+target_link_libraries(viam-camera-realsense PRIVATE
+    grpc++
+    protobuf
```
acmorrow, line 21: Does `viam-camera-realsense` make direct use of `grpc++` and `protobuf`? If not, you can probably drop them. They will likely get automatically re-added because they are `PUBLIC` library dependencies of `viam-cpp-sdk::viamsdk`.

CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    turbojpeg
+)
+
+# The viam C++ SDK 
+find_package(viam-cpp-sdk REQUIRED)
```
acmorrow, line 26: I'd probably move this up above declaring the targets, maybe to around line 9.

CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+target_link_libraries(viam-camera-realsense PRIVATE
+    viam-cpp-sdk::viamsdk
+)
+
+# The realsense library 
+find_package(realsense2 REQUIRED)
```
acmorrow, line 32: I'd also move this up next to the other `find_package` call.

CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+target_link_libraries(viam-camera-realsense PRIVATE
+    realsense2::realsense2
+)
+# Add all the flags
+target_compile_features(viam-camera-realsense PRIVATE 
+    cxx_std_17
+)
+target_compile_options(viam-camera-realsense PRIVATE 
+    -O4
```
acmorrow, line 41: As far as I know, GCC only applies meaning to `-O3` and higher values are just reduced to that. And I'd rarely recommend `-O3` anyway: it can lead to larger code which is actually slower. Consider `-O2`.

CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+target_link_libraries(viam-camera-realsense PRIVATE
+    realsense2::realsense2
+)
+# Add all the flags
+target_compile_features(viam-camera-realsense PRIVATE 
+    cxx_std_17
+)
+target_compile_options(viam-camera-realsense PRIVATE 
+    -O4 
+    -pthread
```
acmorrow, line 42: A better way to get the `-pthread` flag is to use the CMake pseudo-library `Threads::Thread`: https://github.com/viamrobotics/viam-cpp-sdk/blob/main/CMakeLists.txt#L211-L213

CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+# Add all the flags
+target_compile_features(viam-camera-realsense PRIVATE 
+    cxx_std_17
+)
+target_compile_options(viam-camera-realsense PRIVATE 
+    -O4 
+    -pthread 
+    -Wredundant-move 
+    -Wpessimizing-move 
+    -Wl,-ldl
```
acmorrow, line 45: CMake has an abstraction for this:
- https://cmake.org/cmake/help/latest/variable/CMAKE_DL_LIBS.html.
- https://github.com/viamrobotics/viam-mlmodelservice-triton/blob/main/src/CMakeLists.txt#L54

CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+set(CMAKE_CXX_STANDARD 17)
+
+# The image encoders
+set(THIRD_PARTY_SOURCES src/third_party/fpng.cpp src/third_party/lodepng.cpp)
+
+# construct the executable
+add_executable(viam-camera-realsense src/camera_realsense.cpp ${THIRD_PARTY_SOURCES})
+
+enable_testing()
+add_subdirectory(tests)
```
acmorrow, line 17: I'd move this down to the bottom after declaring your targets.

Makefile
```diff
@@ -1,27 +1,18 @@
...
+format: src/*.cpp src/*.hpp
+	clang-format -i --style="{BasedOnStyle: Google, IndentWidth: 4, ColumnLimit: 100}" src/*.cpp && \
+	clang-format -i --style="{BasedOnStyle: Google, IndentWidth: 4, ColumnLimit: 100}" src/*.hpp
+
+viam-camera-realsense: 
+	rm -rf build/ && \
+	mkdir build && \
+	cd build && \
+	cmake .. && \
+	make && \
```
acmorrow, line 11: Same thought that this is going to get you a serial build by default.

acmorrow (changes requested):
Looks good, lots of little small things and improvements, but nothing super important. I do note one future consideration about how testing will work.

Makefile
```diff
@@ -1,27 +1,18 @@
 # compile the binary
```
bazile-clyde, line 1: Is that comment stale now? It doesn't look like it goes with `format` below.

Makefile
```diff
@@ -1,27 +1,18 @@
...
-SDK_INCLUDE = -I$(SDK_LOCATION)/include -I$(SDK_LOCATION)/include/viam/api -L$(SDK_LOCATION)/lib
-SDK_FLAGS = -lviamsdk -lviam_rust_utils -lviamapi 
-
-LIB_FLAGS = $(shell PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --cflags grpc++ realsense2 --libs protobuf grpc++ libturbojpeg realsense2)
-
-viam-camera-realsense: $(SERVER_TARGETS)
-	$(CPP_COMPILER) -std=c++17 -o viam-camera-realsense camera_realsense.cpp $(THIRD_PARTY_SOURCES) $(SDK_INCLUDE) $(LIB_FLAGS) $(SDK_FLAGS) $(GCC_FLAGS)
+format: src/*.cpp src/*.hpp
+	clang-format -i --style="{BasedOnStyle: Google, IndentWidth: 4, ColumnLimit: 100}" src/*.cpp && \
+	clang-format -i --style="{BasedOnStyle: Google, IndentWidth: 4, ColumnLimit: 100}" src/*.hpp
```
bazile-clyde, line 4: [opt] If you want to do this in one line, you can do...
```
$ ls src/*.{cpp,hpp}
src/camera_realsense.cpp  src/camera_realsense.hpp
```
or...
```
$ ls src/*.cpp src/*.hpp 
src/camera_realsense.cpp  src/camera_realsense.hpp
```

bazile-clyde (approved):
This looks great to me! The README updates making things really clear too. LGTM once Andrew's comments are addressed.

tests/camera_realsense_test.cpp
bhaney: This is such a good point, because I could see situations where we would like to run tests on a real intel realsense (maybe to see if the hardware acts as expected) vs. a mock or simulated data stream (for the CI tests on github). Yes, making the tests into a library, and then creating two separate binaries (one for an actual realsense, one for a fake realsense in the CI) would be great

bhaney: While this is a good idea, I think I'll wait until the next PR to implement, or else it will make this one a bit big

tests/camera_realsense_test.cpp
bhaney: got Rid of the link to GTest::Main

tests/CMakeLists.txt
bhaney: removed

src/camera_realsense.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include <viam/sdk/module/service.hpp>
+#include <viam/sdk/registry/registry.hpp>
+#include <viam/sdk/rpc/server.hpp>
+
+constexpr char kResourceType[] = "CameraRealSense";
+constexpr char kAPINamespace[] = "viam";
+constexpr char kAPIType[] = "camera";
+constexpr char kAPISubtype[] = "realsense";
+
+namespace realsense {
```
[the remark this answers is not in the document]
bhaney: done, created viam::realsense

src/camera_realsense.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include <tuple>
+#include <vector>
+#include <viam/sdk/components/camera/camera.hpp>
+#include <viam/sdk/components/camera/server.hpp>
+#include <viam/sdk/components/component.hpp>
+#include <viam/sdk/module/service.hpp>
+#include <viam/sdk/registry/registry.hpp>
+#include <viam/sdk/rpc/server.hpp>
+
+constexpr char kResourceType[] = "CameraRealSense";
```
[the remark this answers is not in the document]
bhaney: done

CMakeLists.txt
bhaney: done

CMakeLists.txt
bhaney: done

CMakeLists.txt
bhaney: done

Makefile
bhaney: switched to using Ninja

CMakeLists.txt
bhaney: done, moved to the bottom of the CMakeLists

src/camera_realsense.hpp
bhaney: done

.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Create build directory
+        run: mkdir build
+
+      - name: Configure CMake
+        working-directory: build
+        run: cmake ..
+
+      - name: Build the binary
+        working-directory: build
+        run: make
```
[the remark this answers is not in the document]
bhaney: switched to Ninja

CMakeLists.txt
bhaney: removed - you were right

CMakeLists.txt
bhaney: done

CMakeLists.txt
bhaney: done

Makefile
bhaney: nice suggestion, implemented

Makefile
bhaney: changed

tests/camera_realsense_test.cpp
acmorrow: Yeah, that makes sense to me.

acmorrow (approved):
LGTM!

PR #10, opened 2023-10-23
RSDK-3054, RSDK-5389: Create CI tests and integration tests for intel realsense

bhaney:
This PR:
- Creates integration tests in Go that can be downloaded as binaries and run on any (Linux aarch64) computer connected to an intel realsense device. Running the binary can be added to the QA plan
- Turns the viam-camera-realsense in a library called libviamrealsense such that testing in CI can be more easily done on its individual methods. 

The CI tests now have:
- a test validating the realsense camera config

The integration tests do the following:
- Set up the camera with a 640x480 resolution
- Call the "get_images" method
- Shut down the server

As expected, the integrations tests currently fail because the of "Segmentation Fault" that happens on the server shutdown


bazile-clyde (approved):
This all LGTM! I was _not_ able to run this properly on my own setup because of Pi problems FYI but trust you with the testing. Thanks for this!

bhaney:
> This all LGTM! I was _not_ able to run this properly on my own setup because of Pi problems FYI but trust you with the testing. Thanks for this!

Ah that's unfortunate! The integration tests haven't been uploaded yet to the server, but they will be uploaded once this PR is merged.

CMakeLists.txt
```diff
@@ -15,20 +15,37 @@ find_package(viam-cpp-sdk REQUIRED)
 # The image encoders
 set(THIRD_PARTY_SOURCES src/third_party/fpng.cpp src/third_party/lodepng.cpp)
 
-# construct the executable
-add_executable(viam-camera-realsense src/camera_realsense.cpp ${THIRD_PARTY_SOURCES})
+# construct the library
+add_library(viamrealsense src/camera_realsense.cpp ${THIRD_PARTY_SOURCES})
 
-target_link_libraries(viam-camera-realsense PRIVATE
+target_link_libraries(viamrealsense PRIVATE
```
acmorrow, line 21: I think you should tag `viam-cpp-sdk::viamsdk` and `realsense2::realsense2` as `PUBLIC` link libraries for `viamrealsense`. Types from those libraries are used in both the interface to this library (the header file) and the implementation, so making it `PUBLIC` makes sense. I think `turbojpeg` should remain `PRIVATE`.

CMakeLists.txt
```diff
@@ -15,20 +15,37 @@ find_package(viam-cpp-sdk REQUIRED)
...
+
+# construct the module executable
+add_executable(viam-camera-realsense main.cpp)
+
+target_link_libraries(viam-camera-realsense PRIVATE
+    viamrealsense
+    viam-cpp-sdk::viamsdk
+    realsense2::realsense2
+    turbojpeg
+    ${CMAKE_DL_LIBS}
```
acmorrow, line 46: I'd expect you can drop all the library dependencies here except `viamrealsense` once you make them `PUBLIC` on `viamrealsense` per my above comment.

test/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ find_package(GTest REQUIRED)
 add_executable(realsense_tests camera_realsense_test.cpp)
 
 include_directories(${CMAKE_SOURCE_DIR}/src)
-target_link_libraries(realsense_tests GTest::GTest viam-cpp-sdk::viamsdk)
+target_link_libraries(realsense_tests 
+    viamrealsense 
+    GTest::GTest 
+    viam-cpp-sdk::viamsdk
```
acmorrow, line 11: This will also get picked up transitively if you make it a `PUBLIC` dependency of `viamrealsense`.

CMakeLists.txt
```diff
@@ -15,20 +15,37 @@ find_package(viam-cpp-sdk REQUIRED)
...
     turbojpeg
     ${CMAKE_DL_LIBS}
 )
 
 # Add all the flags
-target_compile_features(viam-camera-realsense PRIVATE 
+target_compile_features(viamrealsense PRIVATE 
     cxx_std_17
 )
+target_compile_options(viamrealsense PRIVATE
```
acmorrow, line 32: You probably don't really need these on the mainline, but I guess it is also harmless.

CMakeLists.txt
```diff
@@ -15,20 +15,37 @@ find_package(viam-cpp-sdk REQUIRED)
...
+target_link_libraries(viamrealsense PRIVATE
     viam-cpp-sdk::viamsdk
     realsense2::realsense2
     turbojpeg
     ${CMAKE_DL_LIBS}
 )
 
 # Add all the flags
-target_compile_features(viam-camera-realsense PRIVATE 
+target_compile_features(viamrealsense PRIVATE
```
acmorrow, line 29: Similarly, I'd expect that this should be a `PUBLIC` `target_compile_feature` if consumers of `viamrealsense` will require C++17, which I'm fairly sure they do.

src/camera_realsense.cpp
```diff
@@ -901,6 +901,19 @@ std::vector<std::string> validate(sdk::ResourceConfig cfg) {
             }
         }
     }
+    if (attrs->count("sensors") >= 1) {
```
acmorrow, line 904: Commenting here because GH doesn't let you comment on unchanged lines.

There are several functions and constants that are local to this `.cpp`, meaning that they aren't declared in the header. For example `encodeColorPNG`. I'd recommend putting all of those functions and constants that don't form part of the interface inside the unnamed namespace (e.g. `namespace { ... }  // namespace`). Doing so ensures that those functions have internal linkage in the library.

acmorrow (approved):
I think there are some simplifications that can be made by allowing library and flag things to be `PUBLIC` and transitive, and I think some things in the library should be pushed to have internal linkage. Those are all small things that you can do or not, none of them are correctness issues. I think the split you have made looks fine in general, so LGTM for CMake stuff. I did not review the tests or appimage changes since it looks like Clyde was already happy with those.

CMakeLists.txt
bhaney: done

CMakeLists.txt
bhaney: done

test/CMakeLists.txt
bhaney: done

CMakeLists.txt
bhaney: done

src/camera_realsense.cpp
bhaney: put all the other functions in the unnamed namespace

PR #16, opened 2023-11-07
Update README.md

bhaney:
Add a table of known supported hardware


bazile-clyde (approved):
Thanks for hopping on this so quickly!

PR #19, opened 2023-11-29
RSDK-5869: Destroy compression handle for both success and failure cases.

dgottlieb:


bhaney (approved):
Glad it was such a simple fix!

PR #22, opened 2023-12-29
Remove errant comma in json attributes

randhid:
Remove the trailing comma after the last element in the json example attributes.


sguequierre (approved):
Thank you!

PR #27, opened 2024-02-12
RSDK-4281: Shutting down viam-server causes intel realsense module to seg-fault if the component is currently running

hexbabe:
https://viam.atlassian.net/browse/RSDK-4281

## Context for why this solution (hopefully) works

I built the module as a pure binary and tried repro-ing as per the ticket. Could not get it to repro after many attempts. Figured it might be an appimage issue. AppImages are still pretty black-box to me. Did the most sane thing 😂 and just bumped our dependency to a more recent version. Issue seems to be gone: tested on my RPI and Ubuntu setups

Further context: I analyzed a coredump of the crash, and this is what I saw in the stack and threads
image. `libgio` appears. The only thing I can think of that might depend on this is the AppImage stuff. Adding @abe-winter to fact check this.

Further further context: I know on the ticket it was mentioned that this bug is only reproducible when the stream is on. I `gdb attach`ed to the binary process during termination and the result was:
```
Thread 25 "grpcpp_sync_ser" received signal SIGSEGV, Segmentation fault.
[Switching to LWP 24522]
0x0000ffffb9531ea4 in pthread_mutex_lock () from /tmp/.mount_viam-c6axjgw/runtime/compat/lib/aarch64-linux-gnu/libc.so.6
```
My guess is the AppImage and the grpcpp dep aren't playing well together and are not freeing memory + terminating properly?

## Testing
Could someone also test on their machine(s) to verify?

[Download AppImage](https://drive.google.com/file/d/1gLZH0az6N7V_XZTTIshfsk5LtCfOhMim/view?usp=sharing)

### DIFF
--- a/packaging/appimages/viam-camera-realsense-aarch64.yml
+++ b/packaging/appimages/viam-camera-realsense-aarch64.yml
@@ -44,4 +44,5 @@ AppDir:
         AIX_TARGET: usr/bin/viam-camera-realsense
 AppImage:
   arch: aarch64
+  comp: gzip
   update-information: zsync|http://packages.viam.com/apps/camera-servers/viam-camera-realsense-${TAG_NAME}-aarch64.AppImage.zsync
--- a/src/camera_realsense.cpp
+++ b/src/camera_realsense.cpp
@@ -460,7 +460,6 @@ CameraRealSense::~CameraRealSense() {
     std::unique_lock<std::mutex> lock(this->device_->mutex);
     this->device_->shouldRun = false;
     this->device_->cv.wait(lock, [this] { return !(device_->isRunning); });
-    std::this_thread::sleep_for(std::chrono::milliseconds(100));
 }
 
 void CameraRealSense::reconfigure(sdk::Dependencies deps, sdk::ResourceConfig cfg) {
--- a/src/camera_realsense.cpp
+++ b/src/camera_realsense.cpp
@@ -460,6 +460,7 @@ CameraRealSense::~CameraRealSense() {
     std::unique_lock<std::mutex> lock(this->device_->mutex);
     this->device_->shouldRun = false;
     this->device_->cv.wait(lock, [this] { return !(device_->isRunning); });
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
 }
 
 void CameraRealSense::reconfigure(sdk::Dependencies deps, sdk::ResourceConfig cfg) {
--- a/etc/Dockerfile.cuda.debian.bookworm
+++ b/etc/Dockerfile.cuda.debian.bookworm
@@ -93,7 +93,7 @@ RUN apt install -y \
 RUN pip3 install -U pip setuptools urllib3==1.26.12 requests==2.26.0 --break-system-packages
 
 # install appimage-builder
-RUN pip3 install --break-system-packages git+https://github.com/AppImageCrafters/appimage-builder.git@f38699ef3644fa5409a5a262b7b6d99d6fb85db9
+RUN pip3 install --break-system-packages git+https://github.com/AppImageCrafters/appimage-builder.git@61c8ddde9ef44b85d7444bbe79d80b44a6a5576d
 
 # install Go
 RUN apt install -y golang-go 
--- a/etc/Dockerfile.debian.bookworm
+++ b/etc/Dockerfile.debian.bookworm
@@ -88,7 +88,7 @@ RUN apt install -y \
 RUN pip3 install -U pip setuptools urllib3==1.26.12 requests==2.26.0 --break-system-packages
 
 # install appimage-builder
-RUN pip3 install --break-system-packages git+https://github.com/AppImageCrafters/appimage-builder.git@f38699ef3644fa5409a5a262b7b6d99d6fb85db9
+RUN pip3 install --break-system-packages git+https://github.com/AppImageCrafters/appimage-builder.git@61c8ddde9ef44b85d7444bbe79d80b44a6a5576d
 
 # install Go
 RUN apt install -y golang-go 


packaging/appimages/viam-camera-realsense-aarch64.yml
```diff
@@ -44,4 +44,5 @@ AppDir:
         AIX_TARGET: usr/bin/viam-camera-realsense
 AppImage:
   arch: aarch64
+  comp: gzip
```
hexbabe, line 47: Required now by the bumped version of appimage-builder. Let me know if we should use a different compression method

seanavery (changes requested):
Tested out appimage on RPI4 bullseye, unfortunately, it appears I am still getting segfault. Here are the logs:

```
2024-02-12T19:10:05.710Z   error robot_server   server/entrypoint.go:166   Fatal error running server, exiting now: error closing module manager: error while stopping module intel: signal: segmentation fault  
2024-02-12T19:10:05.709Z   error robot_server   server/entrypoint.go:186   error serving web   error error closing module manager: error while stopping module intel: signal: segmentation fault  
2024-02-12T19:10:05.608Z   info robot_server.process.intel_/home/viam/realsense-img.StdOut   pexec/managed_process.go:244   \_ [2024-02-12 14:10:05.608740] [0x0000007fa9f2d040] [info] Shutting down gracefully.  
2024-02-12T19:10:05.608Z   info robot_server.process.intel_/home/viam/realsense-img.StdOut   pexec/managed_process.go:244   \_ end of serve func  
2024-02-12T19:10:05.607Z   info robot_server.process.intel_/home/viam/realsense-img   pexec/managed_process_unix.go:91   stopping process 180603 with signal terminated  
2024-02-12T19:10:05.607Z   info robot_server.process.intel_/home/viam/realsense-img.StdOut   pexec/managed_process.go:244   \_ end of frameLoop function  
2024-02-12T19:10:05.606Z   info robot_server.process.intel_/home/viam/realsense-img.StdOut   pexec/managed_process.go:244   \_ end of frameLoop loop  
2024-02-12T19:10:05.606Z   info robot_server.process.intel_/home/viam/realsense-img.StdOut   pexec/managed_process.go:244   \_ [frameLoop] pipeline stopped, exiting frame loop  
2024-02-12T19:10:04.977Z   info robot_server.process.intel_/home/viam/realsense-img.StdOut   pexec/managed_process.go:244   \_ [frameLoop] start loop  
2024-02-12T19:10:04.801Z   info robot_server.process.intel_/home/viam/realsense-img.StdOut   pexec/managed_process.go:244   \_ [frameLoop] start loop  
2024-02-12T19:10:04.767Z   info robot_server   rpc/server.go:816   stopped cleanly 
```

abe-winter (commented):
other people on this review understand appimage + realsense better than I do so will let them do signoff

from my perspective: no harm in bumping appimage builder even if it doesn't solve the problem. would prefer clearer repro case, but this is a thorny one

hexbabe:
> other people on this review understand appimage + realsense better than I do so will let them do signoff
> 
> from my perspective: no harm in bumping appimage builder even if it doesn't solve the problem. would prefer clearer repro case, but this is a thorny one

Yeah... @seanavery and I were investigating a little offline and upon trying to repro a second time on the same setup (RPI4 Bullseye), the segfault didn't happen. I tried the pure binary method to try to be able to find the exact cause of it, but seeing that the error only happens with AppImages, it makes it extremely difficult to do so using debuggers.

src/camera_realsense.cpp
```diff
@@ -460,6 +460,7 @@ CameraRealSense::~CameraRealSense() {
     std::unique_lock<std::mutex> lock(this->device_->mutex);
     this->device_->shouldRun = false;
     this->device_->cv.wait(lock, [this] { return !(device_->isRunning); });
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
hexbabe, line 463: I didn't want this to be part of the solution we implement, but it does get the job done. Not exactly sure why it prevents segfault

hexbabe:
Synced with Benji and we think a good course of action to take is to implement these bandaid fixes, but also hand off a minimal example of the segfault with a reduced version of the module to the SDK team, since the root of the segfault more than likely comes from our C++ SDK.

src/camera_realsense.cpp
seanavery: @hexbabe Appimage seems to be working without segfault. Although as mentioned, it is hard to repro.

I am ok with adding timeout on deconstuctor. Would be good to add TODO comment here with Jira ticket linking to SDK investigation.

seanavery (approved):

src/camera_realsense.cpp
bazile-clyde: > Not exactly sure why it prevents the segfault.
So generally, you do _not_ want to do this. We don't know if this solves the problem or makes the problem less likely to happen since we don't know what the root cause of the problem is. We could be setting ourselves up for failure by making this harder to reproduce. In C++, you don't actually call destructors explicitly. This could cause non-obvious 100ms delays in downstream functions. If you're wondering why this seems to fix the problem, you can see what happens just before and after this destructor is called, possibly by printing out a stack trace. _Something_ is using memory it's not supposed to. We'd want to find what that something is.

hexbabe: > Synced with Benji and we think a good course of action to take is to implement these bandaid fixes, but also hand off a minimal example of the segfault with a reduced version of the module to the SDK team, since the root of the segfault more than likely comes from our C++ SDK.

@bazile-clyde would you suggest that we don't go push the timeout even as a temporary fix? I am working on a version of the module that will make it easy to repro

bazile-clyde: Yes, I think it's better to wait for a more permanent, clear fix. This may introduce more complicated errors around object destruction.

hexbabe: Okay that sounds good to me. I will continue making the reduced module and I think it's fine to bump the appimage-builder dependency.

PR #37, opened 2024-04-18
RSDK-7243   (FIX) Polish hacky build for Realsense linux/amd64 and make it official CI to publish to registry

hexbabe:
the `appimage-tag.yml` flow (in charge of building and publishing) is breaking on main. I should've tested this sorry


randhid:
@hexbabe bijan is on pto for a week.

randhid (approved):

PR #48, opened 2024-07-15
[RSDK-8195]   Set `supports_pcd` to true in RealSense

hexbabe:
[RSDK-8195](https://viam.atlassian.net/browse/RSDK-8195)

Forgot to change properties after adding new pcd method

Tested with robot usage of RealSense as a dependency for the join-point-clouds module to [verify that we can invoke get_properties properly and that pcd support is now true](https://github.com/viam-modules/join-point-clouds/blob/a473eac1b52801f34492936746017e00f2e140ac/src/main.cpp#L48)

[RSDK-8195]: https://viam.atlassian.net/browse/RSDK-8195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/camera_realsense.cpp
+++ b/src/camera_realsense.cpp
@@ -77,15 +77,6 @@ std::tuple<RealSenseProperties, bool, bool> CameraRealSense::initialize(sdk::Res
             littleEndianDepth = endian_bool;
         }
     }
-    bool enablePointClouds = false;
-    if (attrs->count("enable_point_clouds") == 1) {
-        std::shared_ptr<sdk::ProtoType> pointclouds_proto = attrs->at("enable_point_clouds");
-        auto pointclouds_value = pointclouds_proto->proto_value();
-        if (pointclouds_value.has_bool_value()) {
-            bool pointclouds_bool = static_cast<bool>(pointclouds_value.bool_value());
-            enablePointClouds = pointclouds_bool;
-        }
-    }
     bool disableDepth = true;
     bool disableColor = true;
     std::vector<std::string> sensors;
@@ -132,9 +123,6 @@ std::tuple<RealSenseProperties, bool, bool> CameraRealSense::initialize(sdk::Res
         std::cout << std::boolalpha << "depth little endian encoded: " << littleEndianDepth
                   << std::endl;
     }
-    props.enablePointClouds = enablePointClouds;
-    std::string pointcloudString = (enablePointClouds) ? "true" : "false";
-    std::cout << "point clouds enabled: " << pointcloudString << std::endl;
     std::promise<void> ready;
     std::thread cameraThread(frameLoop, pipe, ref(ready), device_, props.depthScaleMm);
     std::cout << "waiting for camera frame loop thread to be ready..." << std::endl;
--- a/src/camera_realsense.cpp
+++ b/src/camera_realsense.cpp
@@ -238,8 +238,8 @@ sdk::Camera::raw_image CameraRealSense::get_image(std::string mime_type,
 }
 
 sdk::Camera::properties CameraRealSense::get_properties() {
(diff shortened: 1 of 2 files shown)


hexbabe:
@bhaney since Rand is OOO

bhaney (changes requested):
There is a config parameter in the realsense module called "enable_point_clouds" which would toggle on or off the `enablePointClouds ` bool. You should go through the code and remove the config parameter, as well as the enablePointClouds bool 


struct RealSenseProperties also has enablePointClouds bool as a field.

And please continue adding me to PRs for the intel realsense, even when Rand is around

hexbabe:
> There is a config parameter in the realsense module called "enable_point_clouds" which would toggle on or off the `enablePointClouds ` bool. You should go through the code and remove the config parameter, as well as the enablePointClouds bool
> 
> struct RealSenseProperties also has enablePointClouds bool as a field.
> 
> And please continue adding me to PRs for the intel realsense, even when Rand is around

Got it. I removed `enable_point_clouds` in config validation and traces of `enablePointClouds` from the hpp and cpp files

randhid:
> > There is a config parameter in the realsense module called "enable_point_clouds" which would toggle on or off the `enablePointClouds ` bool. You should go through the code and remove the config parameter, as well as the enablePointClouds bool
> > struct RealSenseProperties also has enablePointClouds bool as a field.
> > And please continue adding me to PRs for the intel realsense, even when Rand is around
> 
> Got it. I removed `enable_point_clouds` in config validation and traces of `enablePointClouds` from the hpp and cpp files

Justy FYI - removing configuration attributes of a model is a breaking change.

I do not think we're at enough use of this mode to warrant a migration, but for something like webcam or transform camera I would have insisted we investigate usage and run a migration script to remove the attribute.

But for here - please update the attribute in the readme if it exists.

hexbabe:
> > > There is a config parameter in the realsense module called "enable_point_clouds" which would toggle on or off the `enablePointClouds ` bool. You should go through the code and remove the config parameter, as well as the enablePointClouds bool
> > > struct RealSenseProperties also has enablePointClouds bool as a field.
> > > And please continue adding me to PRs for the intel realsense, even when Rand is around
> > 
> > 
> > Got it. I removed `enable_point_clouds` in config validation and traces of `enablePointClouds` from the hpp and cpp files
> 
> Justy FYI - removing configuration attributes of a model is a breaking change.
> 
> I do not think we're at enough use of this mode to warrant a migration, but for something like webcam or transform camera I would have insisted we investigate usage and run a migration script to remove the attribute.
> 
> But for here - please update the attribute in the readme if it exists.

I don't think this attribute is documented

randhid:
> > > > There is a config parameter in the realsense module called "enable_point_clouds" which would toggle on or off the `enablePointClouds ` bool. You should go through the code and remove the config parameter, as well as the enablePointClouds bool
> > > > struct RealSenseProperties also has enablePointClouds bool as a field.
> > > > And please continue adding me to PRs for the intel realsense, even when Rand is around
> > > 
> > > 
> > > Got it. I removed `enable_point_clouds` in config validation and traces of `enablePointClouds` from the hpp and cpp files
> > 
> > 
> > Justy FYI - removing configuration attributes of a model is a breaking change.
> > I do not think we're at enough use of this mode to warrant a migration, but for something like webcam or transform camera I would have insisted we investigate usage and run a migration script to remove the attribute.
> > But for here - please update the attribute in the readme if it exists.
> 
> I don't think this attribute is documented in the readme or in docs

perfect

bhaney (approved):
LGTM

PR #50, opened 2024-08-13
fix sentence which described where the image is

nfranczak:


hexbabe (approved):
ngl I thought you were gonna put the image below instead, but this works too

nfranczak:
Haha I can do that instead no biggie

PR #52, opened 2024-08-27
Adding Service Account for Workload Identity

etai-shuchatowitz:
Switch from repo secret to WIF for GCP auth

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -40,10 +40,13 @@ jobs:
     - name: Build Integration Tests
       run: make ${{ matrix.make_target }}
 
-    - name: Authorize GCP Upload
-      uses: google-github-actions/auth@v1
+    - id: 'auth'
+      name: 'Authenticate to Google Cloud'
+      uses: 'google-github-actions/auth@v1'
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -40,10 +40,13 @@ jobs:
...
 
-    - name: Authorize GCP Upload
-      uses: google-github-actions/auth@v1
+    - id: 'auth'
+      name: 'Authenticate to Google Cloud'
+      uses: 'google-github-actions/auth@v1'
       with:
-        credentials_json: '${{ secrets.GCP_CREDENTIALS }}'
+        create_credentials_file: true
+        workload_identity_provider: 'projects/385154741571/locations/global/workloadIdentityPools/viam-app-id/providers/github-provider-app'
```
hexbabe, line 48: does this also need to be a var?

etai-shuchatowitz: If you're using it in multiple places, I would make it a var. These aren't secret so the only thing is to allow for a change once, apply everywhere approach

abe-winter (approved):

PR #54, opened 2024-12-04
Tested D435 works on Pi4B bookworm

seanavery:


hexbabe:
Should we merge this?

seanavery:
@hexbabe  I think so.

We should be able to fill this table up more if we get `libuvc` option merged :)

PR #60, opened 2025-05-21
starting to clean build

erh:


hexbabe (approved):
Nice!! Much better Makefile idioms

PR #65, opened 2025-05-23
Protect device querying

hexbabe:


seanavery (approved):
LGTM!

Out of curiosity, what bad behavior or errors were we getting with concurrent device querying?

hexbabe:
> LGTM!
> 
> 
> 
> Out of curiosity, what bad behavior or errors were we getting with concurrent device querying?

Really quite bad. The camera would hang on reconfigure and I would only see the error in dmesg, segfault in librealsense.so. Since it was only caught in dmesg I assumed it was a kernel (usb query) interaction and I think I am right

src/camera_realsense.cpp
```diff
@@ -631,7 +653,7 @@ void on_device_reconnect(rs2::event_information& info, rs2::pipeline pipeline,
     RealSenseProperties props;
     try {
         std::tie(pipeline, props) =
-            startPipeline(device->disableDepth, device->depthWidth, device->depthHeight,
+            startPipeline(*(device->context), device->disableDepth, device->depthWidth, device->depthHeight,
```
nicksanford, line 656: why are you dereferencing the `device->context` pointer here?

src/camera_realsense.hpp
```diff
@@ -47,10 +47,13 @@ struct DeviceProperties {
...
     std::condition_variable cv;
     std::mutex mutex;
+    rs2::context* context;
+
     // DeviceProperties constructor
     DeviceProperties(int colorWidth_, int colorHeight_, bool disableColor_, int depthWidth_,
                      int depthHeight_, bool disableDepth_,
-                     AtomicFrameSet& frames_ref)
+                     AtomicFrameSet& frames_ref,
+                     rs2::context* ctx)
```
nicksanford, line 56: why are we receiving a raw pointer to ctx? Where does the memory that the pointer points to get allocated? Don't we need to free ctx at some point? If so, when does that happen?

src/camera_realsense.cpp
```diff
@@ -351,9 +354,11 @@ void frameLoop(rs2::pipeline pipeline, std::promise<void>& ready,
     }
     // start the callback function that will look for camera disconnects and reconnects.
     // on reconnects, it will close and restart the pipeline and thread.
-    rs2::context ctx;
-    ctx.set_devices_changed_callback(
-        [&](rs2::event_information& info) { on_device_reconnect(info, pipeline, deviceProps); });
+    {
+        std::lock_guard<std::mutex> lock(device_query_mutex);
+        deviceProps->context->set_devices_changed_callback(
+            [&](rs2::event_information& info) { on_device_reconnect(info, pipeline, deviceProps); });
```
seanavery, line 360: Have we verified that this callback successfully reconnects? Do we need a busy loop to keep retrying?

src/camera_realsense.hpp
seanavery: Does the rs2 lib handle the ctx lifecycle?

seanavery (commented):
Do you think it is necessary to use valgrind to check memory management?

src/camera_realsense.cpp
hexbabe: yes that is part of the manual tests. I'll post details tomorrow! thanks for checking

hexbabe:
> Do you think it is necessary to use valgrind to check memory management?

I think the flow for this module is to use ASAN, which I will include in the manual tests.

nicksanford:
> > Do you think it is necessary to use valgrind to check memory management?
> 
> 
> 
> I think the flow for this module is to use ASAN, which I will include in the manual tests.

I agree. Asan is sufficient

nicksanford (approved):
I'm approving this as it fixes a known segfault. I'm concerned that our locking is still wrong and may risk deadlocks. 

I haven't had enough time to go through this code (which I'm not familiar with) to be confident it's free of deadlocks so once this is merged, if there is nothing more urgent to work on, please start going through this code and confirm that in any case where you take a lock while holding another lock it is indeed impossible for a deadlock to occur.

src/camera_realsense.cpp
```diff
@@ -146,15 +181,23 @@ CameraRealSense::CameraRealSense(sdk::Dependencies deps, sdk::ResourceConfig cfg
...
+    VIAM_SDK_LOG(debug) << "destroying realsense camera resource";
     // stop and wait for the frameLoop thread to exit
     if (!this->device_) return;
-    // wait until frameLoop is stopped
-    std::unique_lock<std::mutex> lock(this->device_->mutex);
-    this->device_->shouldRun = false;
-    this->device_->cv.wait(lock, [this] { return !(device_->isRunning); });
+    {
+        // wait until frameLoop is stopped
+        std::unique_lock<std::mutex> lock(this->device_->mutex);
```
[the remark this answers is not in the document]
seanavery: Nvm, I see that is passed into `std::condition_variable cv;`

src/camera_realsense.cpp
```diff
@@ -582,45 +647,99 @@ std::tuple<rs2::pipeline, RealSenseProperties> startPipeline(bool disableDepth,
...
+
+void on_device_reconnect(rs2::event_information& info, std::shared_ptr<DeviceProperties> device) {
+    if (device == nullptr) {
+        VIAM_SDK_LOG(error) << "[on_device_reconnect] received null device properties.";
         return;
     }
 
-    VIAM_SDK_LOG(info) << "[on_device_reconnect] Device was reconnected, restarting pipeline";
+    {
+        std::lock_guard<std::mutex> device_lock(device->mutex);
```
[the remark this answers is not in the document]
hexbabe: comment is wrong. Let me fix

src/camera_realsense.cpp
```diff
@@ -71,7 +104,7 @@ std::tuple<RealSenseProperties, bool, bool> CameraRealSense::initialize(sdk::Res
     bool littleEndianDepth = false;
     if (attrs.count("little_endian_depth")) {
         if (const bool* endian_depth = attrs["little_endian_depth"].get<bool>()) {
-            littleEndianDepth = endian_depth;
+            littleEndianDepth = *endian_depth;
```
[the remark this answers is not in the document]
lia-viam: that looks like a typo from my PR #59, apologies! Per the line above `endian_depth` is a pointer to a `const bool`, so `littleEndianDepth = endian_depth` was setting `littleEndianDepth` to the boolean value of the pointer (`static_cast<bool>(endian_depth)`) rather than the stored boolean value `*endian_depth`

nicksanford (approved):

PR #70, opened 2025-06-12
pin-intel-realsense-sdk-version

nicksanford:
1. Pin intel realsense sdk version to https://github.com/IntelRealSense/librealsense/releases/tag/v2.55.1
2. log intel realsense sdk version the module was built with on startup 
3. log warning if module detects intel realsense using usb 2.x 
4. log warning if module detects intel realsense using fimware below recommended version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ set(CMAKE_CXX_STANDARD 17)
 
 # Define an option for enabling build with AddressSanitizer
 option(ENABLE_SANITIZER "Enable AddressSanitizer" OFF)
+option(CMAKE_BUILD_TYPE RelWithDebInfo)
 
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ find_package(Threads REQUIRED)
 find_package(realsense2 REQUIRED)
 find_package(viam-cpp-sdk REQUIRED)
 
-set(HELPER_SOURCES src/encoding.hpp)
+set(HELPER_SOURCES src/encoding.hpp src/version_helpers.hpp)
 set(THIRD_PARTY_SOURCES src/third_party/fpng.cpp src/third_party/lodepng.cpp) 
 # construct the library
 add_library(viamrealsense src/camera_realsense.cpp ${THIRD_PARTY_SOURCES} ${HELPER_SOURCES}) 
--- a/main.cpp
+++ b/main.cpp
@@ -2,15 +2,17 @@
 
 #include <viam/sdk/common/instance.hpp>
 
-int main(int argc, char** argv) {
-    const std::string usage = "usage: camera_realsense /path/to/unix/socket";
+int main(int argc, char **argv) {
+  const std::string usage = "usage: camera_realsense /path/to/unix/socket";
+  std::cout << "intel realsense SDK version: " << RS2_API_FULL_VERSION_STR
+            << "\n";
 
-    if (argc < 2) {
-        std::cout << "ERROR: insufficient arguments\n";
-        std::cout << usage << "\n";
-        return EXIT_FAILURE;
-    }
+  if (argc < 2) {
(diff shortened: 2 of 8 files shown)


src/camera_realsense.cpp
```diff
@@ -804,13 +1069,17 @@ void on_device_reconnect(rs2::event_information &info, std::shared_ptr<DevicePro
     RealSenseProperties props;
     try {
         rs2::pipeline new_pipeline;
-        std::tie(new_pipeline, props) = startPipeline(device, device->active_serial_number);
+        {
+            std::lock_guard<std::mutex> lock(g_realsense_module_lock);
+            std::tie(new_pipeline, props) = startPipeline(device, device->active_serial_number);
```
nicksanford, line 1074: need this mutex to protect rs2_ctx

src/camera_realsense.cpp
```diff
@@ -60,6 +222,9 @@ static void deregister_device(const std::string &serial_number) {
 
 // initialize will use the ResourceConfigs to begin the realsense pipeline.
 RealSenseProperties CameraRealSense::initialize(sdk::ResourceConfig cfg) {
+    if (debug_enabled) {
```
hexbabe, line 225: Why are we using `debug_enabled`? In my head I was thinking we should do the opposite i.e. remove `debug_enabled` and just use `VIAM_SDK_LOG(debug)`

CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ set(CMAKE_CXX_STANDARD 17)
 
 # Define an option for enabling build with AddressSanitizer
 option(ENABLE_SANITIZER "Enable AddressSanitizer" OFF)
+option(CMAKE_BUILD_TYPE RelWithDebInfo)
```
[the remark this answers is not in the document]
nicksanford: It won't. It just adds debug symbols.

main.cpp
```diff
@@ -1,16 +1,19 @@
 #include "src/camera_realsense.hpp"
 
+#include <memory>
 #include <viam/sdk/common/instance.hpp>
 
-int main(int argc, char** argv) {
-    const std::string usage = "usage: camera_realsense /path/to/unix/socket";
+int main(int argc, char **argv) {
+  const std::string usage = "usage: camera_realsense /path/to/unix/socket";
+  std::cout << "intel realsense SDK version: " << RS2_API_FULL_VERSION_STR
```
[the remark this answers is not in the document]
nicksanford: https://github.com/IntelRealSense/librealsense/blob/master/include/librealsense2/rs.h#L45

src/camera_realsense.cpp
nicksanford: I'm adding support for enabling debug level logs for the intel realsense SDK. That SDK doesn't know bout the viam log level. Also, right now it's not possible (to my knowledge) to have one viam cpp sdk resource at one log level and another at another log level.

As a result I'm hacking around the lack of the cpp sdk feature by detecting the log level of the entire module & enabling more logging if the module is booted at debug level. We also want those logs if a user adds "debug=true" to the realsense config, which won't change the module logger's log level.

I agree this is messy and ugly but I don't see another way to get the logging behavior we need (which includes non viam loggers) without this type of a hack given the current state of the viam cpp sdk.

Let me know if there is something I'm missing.

hexbabe: No with that context this makes sense. Thanks

src/camera_realsense.cpp
```diff
@@ -819,9 +1088,11 @@ void on_device_reconnect(rs2::event_information &info, std::shared_ptr<DevicePro
     // Start the camera std::thread
     std::promise<void> ready;
     device->shouldRun.store(true);
+    VIAM_SDK_LOG(info) << "[on_device_reconnect] device->isRunning.store(true): "
+                       << device->active_serial_number;
     device->isRunning.store(true);
     std::thread cameraThread(frameLoop, std::ref(ready), device, props.depthScaleMm,
-                             std::ref(device->atomic_frame_set));
+                             std::ref(device->atomic_frame_set), module_level_debug.load());
```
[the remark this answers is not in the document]
nicksanford: done

PR #75, opened 2025-07-08
RSDK-11197   Add the correct geometries to the realsense module

hexbabe:
[RSDK-11197](https://viam.atlassian.net/browse/RSDK-11197)

### DIFF
--- a/src/camera_realsense.cpp
+++ b/src/camera_realsense.cpp
@@ -15,6 +15,7 @@
 #include <viam/sdk/module/service.hpp>
 #include <viam/sdk/registry/registry.hpp>
 #include <viam/sdk/rpc/server.hpp>
+#include <viam/sdk/spatialmath/geometry.hpp>
 
 #include "encoding.hpp"
 #include "version_helpers.hpp"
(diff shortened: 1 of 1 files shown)


src/camera_realsense.cpp
```diff
@@ -481,8 +482,7 @@ sdk::Camera::point_cloud CameraRealSense::get_point_cloud(std::string mime_type,
 }
 
 std::vector<sdk::GeometryConfig> CameraRealSense::get_geometries(const sdk::ProtoStruct &extra) {
-    VIAM_SDK_LOG(error) << "get_geometries not implemented";
-    return std::vector<sdk::GeometryConfig>{};
+    return {sdk::GeometryConfig(sdk::pose{-17.5, 0, -12.5}, sdk::box({90, 25, 25}), "box")};
```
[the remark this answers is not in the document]
hexbabe: Good call

randhid:
The geometry looks super correct to me! It's where I'd expect it to be when parented to the arm and the offsets look correct!

randhid (approved):

src/camera_realsense.cpp
```diff
@@ -481,8 +482,9 @@ sdk::Camera::point_cloud CameraRealSense::get_point_cloud(std::string mime_type,
 }
 
 std::vector<sdk::GeometryConfig> CameraRealSense::get_geometries(const sdk::ProtoStruct &extra) {
-    VIAM_SDK_LOG(error) << "get_geometries not implemented";
-    return std::vector<sdk::GeometryConfig>{};
+    // This is the geometry for the D435 and D435i, the only models that we currently support.
+    // See https://github.com/viam-modules/viam-camera-realsense/pull/75 for explanation of values.
```
nicksanford, line 486: Please make a ticket as part of the realsense priductionization epic that reports an error and refuses to construct the camera if the camera model is not the 435 or 435i.

Also please add a comment that the geometries needs to be updated if we start supporting more realsense camera models.

nicksanford (approved):

src/camera_realsense.cpp
hexbabe: Added comment

PR #78, opened 2025-07-29
[RSDK-11481] Implement cli to learn to manage camera

SebastianMunozP:
https://viam.atlassian.net/browse/RSDK-11481

### DIFF
--- a/packaging/appimages/viam-camera-realsense-cli-x86_64.yml
+++ b/packaging/appimages/viam-camera-realsense-cli-x86_64.yml
@@ -30,9 +30,9 @@ AppDir:
     - amd64
     allow_unauthenticated: true
     sources:
-    - sourceline: deb [trusted=yes] http://deb.debian.org/debian bookworm main
-    - sourceline: deb [trusted=yes] http://deb.debian.org/debian-security bookworm-security main
-    - sourceline: deb [trusted=yes] http://deb.debian.org/debian bookworm-updates main
+    - sourceline: deb [trusted=yes] https://deb.debian.org/debian bookworm main
+    - sourceline: deb [trusted=yes] https://deb.debian.org/debian-security bookworm-security main
+    - sourceline: deb [trusted=yes] https://deb.debian.org/debian bookworm-updates main
     include:
     - libprotobuf32:amd64
     - libturbojpeg0:amd64
--- a/packaging/appimages/viam-camera-realsense-cli-x86_64.yml
+++ b/packaging/appimages/viam-camera-realsense-cli-x86_64.yml
@@ -1,0 +1,56 @@
+version: 1
+script:
+ - rm -rf AppDir || true
+ - mkdir -p $TARGET_APPDIR/usr/bin
+ - cp ../../build/viam-camera-realsense-cli $TARGET_APPDIR/usr/bin/
+ - mkdir -p $TARGET_APPDIR/usr/share/icons/viam/256x256/apps/
+ - cp ./viam-server.png $TARGET_APPDIR/usr/share/icons/viam/256x256/apps/viam-server.png
+ - go install github.com/Otterverse/aix@latest
+ - cp `go env GOPATH`/bin/aix $TARGET_APPDIR/usr/bin/
+ - chmod 755 $TARGET_APPDIR/usr/bin/*
+ - |
+  if [ "${SANITIZE}" = "ON" ]; then
+    mkdir -p $TARGET_APPDIR/usr/lib
+    cp /lib/x86_64-linux-gnu/libabsl_* $TARGET_APPDIR/usr/lib/
+    cp /lib/x86_64-linux-gnu/libboost_* $TARGET_APPDIR/usr/lib/
+    cp /lib/x86_64-linux-gnu/libicu* $TARGET_APPDIR/usr/lib/
+  fi
+
+AppDir:
+  path: ./AppDir
+  app_info:
+    id: com.viam.camerarealsense
+    name: viam-camera-realsense-cli
+    icon: viam-server
+    version: "latest"
+    exec: usr/bin/aix
+    exec_args: $@
+  apt:
+    arch:
+    - amd64
+    allow_unauthenticated: true
+    sources:
+    - sourceline: deb [trusted=yes] http://deb.debian.org/debian bookworm main
+    - sourceline: deb [trusted=yes] http://deb.debian.org/debian-security bookworm-security main
+    - sourceline: deb [trusted=yes] http://deb.debian.org/debian bookworm-updates main
+    include:
+    - libprotobuf32:amd64
+    - libturbojpeg0:amd64
+  files:
+    include:
+    - /usr/local/lib/libviamsdk* 
+    - /usr/local/lib/libviamapi* 
+    - /usr/local/lib/libviam_rust_utils* 
+    - /usr/local/lib/librealsense2*
+    exclude:
+    - usr/share/man
+    - usr/share/doc
+  runtime:
+    env:
+        AIX_TARGET: usr/bin/viam-camera-realsense-cli
+        ASAN_OPTIONS: detect_leaks=1 detect_stack_use_after_return=true
+        LSAN_OPTIONS: verbosity=1:log_threads=1
+AppImage:
+  arch: x86_64
+  comp: gzip
+  update-information: zsync|http://packages.viam.com/apps/camera-servers/viam-camera-realsense-cli-latest-x86_64.AppImage.zsync
--- a/src/cli/main.cpp
+++ b/src/cli/main.cpp
@@ -125,12 +125,15 @@ struct my_device
     std::string serial_number;
     std::shared_ptr<rs2::device> device;
     std::shared_ptr<rs2::pipeline> pipe;
-    std::shared_ptr<PointCloudFilter> pointCloudFilter;
+    std::shared_ptr<PointCloudFilter> point_cloud_filter;
     std::shared_ptr<rs2::align> align;
+    std::string output_filename;
+    size_t frame_count = 0;
+    std::chrono::steady_clock::time_point last_report = std::chrono::steady_clock::now();
 };
 
 std::mutex devices_by_serial_mu;
-std::unordered_map<std::string, std::unique_ptr<my_device>> devices_by_serial;
+std::unordered_map<std::string, std::shared_ptr<my_device>> devices_by_serial;
 
 std::mutex frame_set_by_serial_mu;
 std::unordered_map<std::string, std::shared_ptr<rs2::frameset>> frame_set_by_serial;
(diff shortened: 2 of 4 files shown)


src/cli/main.cpp
```diff
@@ -0,0 +1,456 @@
...
+
+struct PointXYZRGB
+{
+    float x, y, z;
+    unsigned int rgb;
+};
+
+bool validPoint(const rs2::vertex &p)
+{
+    const double min_distance = 1e-6;
```
lia-viam, line 20: maybe this was meant to use the `min_distance` defined above?

src/cli/main.cpp
```diff
@@ -0,0 +1,456 @@
...
+    return std::fabs(p.x) >= min_distance || std::fabs(p.y) >= min_distance || std::fabs(p.z) >= min_distance;
+}
+
+std::vector<unsigned char> RGBPointsToPCD(std::tuple<rs2::points, rs2::video_frame> data)
```
lia-viam, line 24: i think this implementation could be mostly copied from the one in my orbbec PR if you like

src/cli/main.cpp
```diff
@@ -0,0 +1,456 @@
...
+        pcdBytes.insert(pcdBytes.end(), rgb, rgb + 4);
+    }
+    return pcdBytes;
+}
+
+class PointCloudFilter
+{
+public:
+    PointCloudFilter() : pointcloud(std::make_shared<rs2::pointcloud>()) {}
+    std::tuple<rs2::points, rs2::video_frame> process(rs2::frameset frameset)
```
lia-viam, line 98: minor but this could be `std::pair` in case it may be more convenient to access it with `pair.first` rather than `std::get<0>(pair)`

src/cli/main.cpp
```diff
@@ -0,0 +1,456 @@
...
+// check if the given stream profiles are matching in terms of width, height and FPS
+// This is used to create a config for software depth-to-color (D2C) alignment
+// It checks if the color and depth profiles have the same width, height, and FPS
+// If they match, it returns true and prints the matching parameters
+// If they do not match, it returns false
+// alignment
+bool checkIfMatchingColorDepthProfiles(const rs2::video_stream_profile &color,
+                                       const rs2::video_stream_profile &depth)
+{
+    if (color.width() == depth.width() &&
```
lia-viam, line 160: minor: such comparisons can be done with `tuple`, eg

```cpp
if (std::tuple(color.width(), color.height(), color.fps()) == std::tuple(depth.width(), depth.height(), depth.fps())) {}
```

lia-viam (commented):
LGTM, just some suggestions/points for discussion

src/cli/main.cpp
SebastianMunozP: Fixed in latest commits

packaging/appimages/viam-camera-realsense-cli-x86_64.yml
```diff
@@ -0,0 +1,56 @@
...
+  fi
+
+AppDir:
+  path: ./AppDir
+  app_info:
+    id: com.viam.camerarealsense
+    name: viam-camera-realsense-cli
+    icon: viam-server
+    version: "latest"
+    exec: usr/bin/aix
```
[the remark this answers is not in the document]
seanavery: Ok, I see the builtin zync hook for updating from storage bucket is present:
```
update-information: zsync|http://packages.viam.com/apps/camera-servers/viam-camera-realsense-cli-latest-x86_64.AppImage.zsync
```

packaging/appimages/viam-camera-realsense-cli-x86_64.yml
```diff
@@ -0,0 +1,56 @@
...
+ - go install github.com/Otterverse/aix@latest
+ - cp `go env GOPATH`/bin/aix $TARGET_APPDIR/usr/bin/
+ - chmod 755 $TARGET_APPDIR/usr/bin/*
+ - |
+  if [ "${SANITIZE}" = "ON" ]; then
+    mkdir -p $TARGET_APPDIR/usr/lib
+    cp /lib/x86_64-linux-gnu/libabsl_* $TARGET_APPDIR/usr/lib/
+    cp /lib/x86_64-linux-gnu/libboost_* $TARGET_APPDIR/usr/lib/
+    cp /lib/x86_64-linux-gnu/libicu* $TARGET_APPDIR/usr/lib/
+  fi
```
hexbabe, line 17: have we tested if ASAN works with this binary?

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+#include <thread>
+#include <map>
+#include <memory>
+#include <mutex>
+#include <unordered_map>
+#include <fstream>
+
+#include <librealsense2/rs.hpp>
+
+static const double min_distance = 1e-6;
```
hexbabe, line 11: unclear why we need to filter min_distance. Is it to reduce noise?

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+
+struct PointXYZRGB
+{
+    float x, y, z;
+    unsigned int rgb;
+};
+
+bool validPoint(const rs2::vertex &p)
```
hexbabe, line 19:
```suggestion
bool isValidPoint(const rs2::vertex &p)
```

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+{
+    return std::fabs(p.x) >= min_distance || std::fabs(p.y) >= min_distance || std::fabs(p.z) >= min_distance;
+}
+
+std::vector<PointXYZRGB> getRGBPoints(std::pair<rs2::points, rs2::video_frame>&& data) {
```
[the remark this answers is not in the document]
hexbabe: Unresolving to ask a question. So does that mean the caller of this function always needs to pass in a std::move value? What if they don't

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+    
+    if (!points || !color) {
+        throw std::runtime_error("Points or color frame data is empty");
+    }
+    
+    int num_points = points.size();
+    const rs2::vertex* vertices = points.get_vertices();
+    const rs2::texture_coordinate* tex_coords = points.get_texture_coordinates();
+    
+    const uint8_t* color_data = reinterpret_cast<const uint8_t*>(color.get_data());
```
hexbabe, line 36: it's my understanding that reinterpret_cast can be dangerous sometimes since it bypasses a lot of compile time safety checks

why is it ok to use here?

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+           << "WIDTH " << pcdPoints.size() << "\n"
+           << "HEIGHT 1\n"
+           << "VIEWPOINT 0 0 0 1 0 0 0\n"
+           << "POINTS " << pcdPoints.size() << "\n"
+           << "DATA binary\n";
+    std::string headerStr = header.str();
+    std::vector<unsigned char> pcdBytes;
+    pcdBytes.insert(pcdBytes.end(), headerStr.begin(), headerStr.end());
+
+    // Assert that PointXYZRGB is a POD type, and that it has no padding, thus can be copied as bytes.
```
hexbabe, line 83: what is a POD type

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+    // Since vector is contiguous, we can just copy the whole thing
```
hexbabe, line 84: how do we know it's continuous? is this obvious through the code?

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+              << "  USB Type Descriptor:            " << dev.get_info(RS2_CAMERA_INFO_USB_TYPE_DESCRIPTOR) << "\n";
+}
+
+// check if the given stream profiles are matching in terms of width, height and FPS
+// This is used to create a config for software depth-to-color (D2C) alignment
+// It checks if the color and depth profiles have the same width, height, and FPS
+// If they match, it returns true and prints the matching parameters
+// If they do not match, it returns false
+// alignment
+bool checkIfMatchingColorDepthProfiles(const rs2::video_stream_profile &color,
```
hexbabe, line 175: This feels like something the SDK should take care of... is there no utility they provide that does this?

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+    my_dev->align = std::make_shared<rs2::align>(RS2_STREAM_COLOR);
+    my_dev->serial_number = serial_number;
+    my_dev->device = dev;
+    my_dev->output_filename = serial_number + "_output.pcd";
+
+    std::cout << "Starting stream for device with serial number: " << my_dev->serial_number << std::endl;
+    auto cfg = createSwD2CAlignConfig(my_dev->pipe, ctx, my_dev->device);
+    if (cfg == nullptr)
+    {
+        std::cerr << "Failed to create matching color-depth config for device: " << my_dev->serial_number << std::endl;
```
hexbabe, line 250: why do we handle the error in this way? why not throw

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+        if (csp.format() != RS2_FORMAT_RGB8)
+        {
+            continue; // Only consider RGB8 format for color
+        }
+        for (auto &dp : depth_profiles)
+        {
+            auto dsp = dp.as<rs2::video_stream_profile>();
+            if (dsp.format() != RS2_FORMAT_Z16)
+            {
+                continue; // Only consider Z16 format for depth
```
hexbabe, line 220: sorry if this is already something we do in the realsense module that is just being mirrored here, but is there a reason why Z16 and RGB8 are the formats we consider valid, as well as matching?

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+        }
+        std::cout << "Device Added: " << dev.get_info(RS2_CAMERA_INFO_SERIAL_NUMBER) << std::endl;
+        printDeviceInfo(dev);
+        startStream(dev.get_info(RS2_CAMERA_INFO_SERIAL_NUMBER), std::make_shared<rs2::device>(dev), ctx);
+    }
+    std::cout << "Device list queried, waiting for device changes..." << std::endl;
+
+    std::cout << "waiting for key press\n";
+    std::cin.get();
+    std::cout << "stopping orbbec program" << std::endl;
```
hexbabe, line 462: realsense

src/cli/main.cpp
```diff
@@ -0,0 +1,471 @@
...
+
+    stopStreams();
+    std::lock_guard<std::mutex> lock(devices_by_serial_mu());
+    devices_by_serial().clear();
+    std::lock_guard<std::mutex> frame_lock(frame_set_by_serial_mu());
+    frame_set_by_serial().clear();
+    std::cout << "stopped orbbec program" << std::endl;
```
hexbabe, line 469: realsense

hexbabe (commented):
Left a mixture of cpp learning questions, regular questions, and suggestions

PR #82, opened 2025-09-10
[RSDK-11884] Updating REALSENSE_VERSION and VIAM_CPP_SDK_TAG in Docke…

SebastianMunozP:
When investigating ways to better control our realsense cameras, found out that there was an option I could not access because the SDK we are using is from April 2024 (RS2_OPTION_REGION_OF_INTEREST). On top of that, there are the usual bug fixes and improvements.

CPP-SDK-19 would allow us to remove the need for specifying an empty viam::sdk::LinkConfig{} in viam::sdk::ResourceConfig, as well as other improvements, including a fix in MacOS Conan.

### DIFF
--- a/etc/Dockerfile.debian.bookworm
+++ b/etc/Dockerfile.debian.bookworm
@@ -77,7 +77,7 @@ RUN apt install -y libgtest-dev
 RUN mkdir -p /root/opt/src
 
 # Install librealsense from source
-ENV REALSENSE_VERSION="v2.55.1"
+ENV REALSENSE_VERSION="v2.56.5"
 RUN cd /root/opt/src && \
     git clone https://github.com/IntelRealSense/librealsense.git && \
     cd librealsense && \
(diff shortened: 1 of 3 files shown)


etc/Dockerfile.debian.bookworm
```diff
@@ -77,7 +77,7 @@ RUN apt install -y libgtest-dev
 RUN mkdir -p /root/opt/src
 
 # Install librealsense from source
-ENV REALSENSE_VERSION="v2.55.1"
+ENV REALSENSE_VERSION="v2.56.5"
```
lia-viam, line 80: will we want to stick with this version for a while? i have an open PR with conan to add 2.55.1, but i would want to change it to 2.56.5 if so

lia-viam:
while doing the docker update for this PR would it be possible to add `libgmock-dev` in addition to `libgtest-dev` in the dockerfiles?

PR #83, opened 2025-09-16
RSDK-10894: Add option for module.tar.gz with no AppImage using CMake and CPack

lia-viam:
Following the pattern of https://github.com/viam-modules/universal-robots/pull/102, this is a PR which incrementally introduces the ability to package the module without AppImage. We add a CMake install step, and use CPack to generate a module.tar.gz. 

This will be of limited use until a subsequent PR introducing a conan build. The conan build will be used to generate a build with all static dependencies, and then the CPack deployment creates the module.tgz that can be uploaded to the registry.

The feature flag is `VIAM_REALSENSE_DISABLE_APPIMAGE`, which is off by default, and in the Makefile this is `DISABLE_APPIMAGE`. See PR comments for annotations.

Note that I have tried to update the Makefile but I think ideally we should move away from using Makefiles on top of CMake. The equivalent logic of doing this in the Makefile is a bit different compared to

```
docker pull ghcr.io/viam-modules/viam-camera-realsense:amd64
docker run -it --entrypoint /bin/bash --platform linux/amd64 <container-hash>
git clone https://github.com/lia-viam/viam-camera-realsense.git
git checkout cmake-install
mkdir build && cd build
cmake .. -G Ninja -DVIAM_REALSENSE_ENABLE_TESTS=OFF -DVIAM_REALSENSE_DISABLE_APPIMAGE=ON
ninja package
```

this generates a `module.tar.gz` in the `build` directory:

```
$ ninja package
[4/5] Run CPack packaging tool...
CPack: Create package using TGZ
CPack: Install projects
CPack: - Install project: viam-camera-realsense []
CPack: Create package
CPack: - package: /viam-camera-realsense/build/module.tar.gz generated.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ docker-amd64-ci:
 TAG_VERSION?=latest
 # Define a function for building AppImages
 define BUILD_APPIMAGE
-	ifeq ($(DISABLE_APPIMAGE),1)
+	ifeq ($(DISABLE_APPIMAGE),0)
     	export TAG_NAME=$(TAG_VERSION); \
     	cd packaging/appimages && \
     	mkdir -p deploy && \


Makefile
```diff
@@ -60,22 +63,30 @@ docker-amd64-ci:
...
+	endif
 endef
 
 # Targets for building AppImages
 appimage-arm64: export OUTPUT_NAME = viam-camera-realsense
 appimage-arm64: export ARCH = aarch64
 appimage-arm64: viam-camera-realsense
 	$(call BUILD_APPIMAGE,$(OUTPUT_NAME),$(ARCH))
+ifeq ($(DISABLE_APPIMAGE), 0)
 	cp ./packaging/appimages/$(OUTPUT_NAME)-*-$(ARCH).AppImage ./packaging/appimages/deploy/
```
lia-viam, line 83: for an appimage disabled build there is simply `build/module.tar.gz`, not sure where we would want to copy it

cmake/module_tgz_install.cmake
```diff
@@ -0,0 +1,40 @@
+if (LINUX)
+   set(_pre_exclude_regex
+        ".*ld-linux.*\\.so.*"
+        ".*libc\\.so.*"
+        ".*libdl\\.so.*"
+        ".*libgcc.*\\.so.*"
+        ".*libm\\.so.*"
+        ".*libpthread\\.so.*"
+        ".*libssl\\.so.*"
```
lia-viam, line 9: per discussion with @acmorrow we may want to also add
- `libz`
- `libstdc++`
and in a future conan build, prevent conan from building zlib and OpenSSL if they are present on the system.

as for `libstdc++` I am not sure, it depends on the level of granularity of packages for which we could build and distribute a `module.tar.gz`

PR #84, opened 2025-09-16
[RSDK-11774] Updating Realsense module with new get_images signature

SebastianMunozP:


src/module/realsense.hpp
```diff
@@ -23,9 +23,12 @@ static constexpr std::uint64_t maxFrameAgeMs =
     1e3; // time until a frame is considered stale, in miliseconds (equal to 1
 static constexpr size_t MAX_GRPC_MESSAGE_SIZE =
     33554432; // 32MB gRPC message size limit
-const std::string kPcdMimeType = "pointcloud/pcd";
 const std::string service_name = "viam_realsense";
 
+const std::string kColorSourceName = "color";
+const std::string kDepthSourceName = "depth";
+const std::string kPcdMimeType = "pointcloud/pcd";
```
hexbabe, line 30: can you confirm this is what it is called in rdk?

hexbabe (approved):
looks great!

src/module/realsense.hpp
SebastianMunozP: pointcloud/pcd is correct

"color" and "depth" for what I can see, they just pass whatever comes from the image:
- [Camera Server]( https://github.com/viamrobotics/viam-cpp-sdk/blob/1ed59dfe43ff1ecaa3899a85f15090a59eb83473/src/viam/sdk/components/private/camera_server.cpp#L93)
- [Camera Client](https://github.com/viamrobotics/viam-cpp-sdk/blob/1ed59dfe43ff1ecaa3899a85f15090a59eb83473/src/viam/sdk/components/private/camera_client.cpp#L141)

PR #88, opened 2025-09-18
Add job for building module.tar.gz and uploading artifact

lia-viam:
This is an intermediate step which will let us test the build changes from #83 

Similar to the dry-run PR for the Orbbec module, this will allow building a module.tar.gz without uploading it to the viam registry, so that we can test it on other machines first


SebastianMunozP (approved):
LGTM

PR #137, opened 2026-04-18
[RSDK-13785] Force libzip to link statically via Conan

SebastianMunozP:
## Summary

- Removes the `pkg_check_modules` / pkg-config path for libzip from `CMakeLists.txt` — always use Conan's `find_package(libzip REQUIRED)` and link statically against `libzip::zip`
- Adds `"libzip/*:shared": False` to `conanfile.py` default options to ensure the Conan-provided libzip is built and linked statically
- Replaces buildjet for GH runners, as for some reason buildjet is not working, its runners never run
- Uploads build artifacts (module.tar.gz) on each PR, no just on each release, this helps test that our CI built modules work OK

## Why

On a fresh Raspberry Pi OS install, `libzip.so.4` is not present. The previous `CMakeLists.txt` preferred the system libzip via pkg-config, which baked in a dynamic `libzip.so.4` dependency at build time even though Conan's libzip was available. This caused the module to fail to start on any machine missing libzip at runtime.

## Test plan

- [x] Build `module.tar.gz` for `linux/arm64` and verify `ldd viam-camera-realsense | grep zip` returns nothing
- [x] Confirm firmware update works (this is the workflow that uses zip to unzip the firmware file)
- [x] Downloaded this module from CI and verified it works OK on amd64, darwin and arm64

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -78,7 +78,7 @@ jobs:
           make module.tar.gz
 
       - name: Upload module artifact
-        if: github.event_name == 'release'
+        if: github.event_name == 'release' || github.event_name == 'pull_request'
         uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.artifact-name }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,43 +58,23 @@ jobs:
             exit 1
           fi
 
-      - name: Build with coverage (Linux amd64 only)
+      - name: Install lcov (Linux amd64 only)
         if: matrix.coverage == true
-        run: |
-          . ./venv/bin/activate
-          GENERATORS="$(pwd)/build-conan/build/Release/generators"
-          mkdir -p build-native && cd build-native
-          cmake .. \
-            -DVIAM_REALSENSE_ENABLE_TESTS=ON \
-            -DVIAM_REALSENSE_ENABLE_COVERAGE=ON \
-            -DCMAKE_BUILD_TYPE=Release \
-            -DCMAKE_TOOLCHAIN_FILE="$GENERATORS/conan_toolchain.cmake" \
-            -DCMAKE_PREFIX_PATH="$GENERATORS"
-          make -j$(nproc)
+        run: sudo apt-get update && sudo apt-get install -y lcov
 
-      - name: Build without coverage (other platforms)
+      - name: Build and run tests with coverage (Linux amd64 only)
+        if: matrix.coverage == true
+        run: make test-coverage
+
+      - name: Build and run tests (other platforms)
         if: matrix.coverage != true
-        run: make build-native
-
-      - name: Run the tests
-        run: |
-          . ./build-conan/build/Release/generators/conanrun.sh
-          cd build-native && ctest --output-on-failure
-
-      - name: Generate coverage report
-        if: matrix.coverage == true
-        run: |
-          sudo apt-get update && sudo apt-get install -y lcov
-          cd build-native
-          lcov --capture --directory . --output-file coverage.info
-          lcov --remove coverage.info '/usr/*' '*/test/*' '*/build-native/_deps/*' --output-file coverage.info
-          lcov --list coverage.info
+        run: make test
 
       - name: Upload coverage reports to Codecov
         if: matrix.coverage == true
         uses: codecov/codecov-action@v5
         with:
-          files: ./build-native/coverage.info
+          files: ./build-conan/coverage.info
           flags: unittests
           # Token is optional for public repos, but recommended for reliability
           # If uploads fail, ensure CODECOV_TOKEN is set in repository secrets
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,12 +15,12 @@ on:
 jobs:
   run-tests:
     name: Run Tests
-    timeout-minutes: 10
+    timeout-minutes: 30
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         include:
-          - os: buildjet-8vcpu-ubuntu-2204-arm
+          - os: github-linux-arm64-8core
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -15,12 +15,12 @@ on:
...
   run-tests:
     name: Run Tests
-    timeout-minutes: 10
+    timeout-minutes: 30
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         include:
-          - os: buildjet-8vcpu-ubuntu-2204-arm
+          - os: github-linux-arm64-8core
```
SebastianMunozP, line 23: buildjet is taking a long time to build and not finishing, replacing it with GH runners as this PR: https://github.com/viam-modules/csi-camera/commit/28455ee712a9ad750abdaa44d22a18347bc8f274

.github/workflows/test.yml
```diff
@@ -40,11 +40,14 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install dependencies
-        if: runner.os == 'macOS'
         run: |
           make setup
           test -f ./venv/bin/activate && . ./venv/bin/activate && conan profile show
-      
+
+      - name: Install Conan dependencies
```
SebastianMunozP, line 47: Now we are going to always us Conan, for all platforms, not just Mac as before. This unifies and simplifies our flow, and in particular, allows us to install libzip through Conan and be usable in our CI as well.

.github/workflows/test.yml
```diff
@@ -55,24 +58,28 @@ jobs:
...
-      - name: Install Conan dependencies
-        if: runner.os == 'macOS'
-        run: |
-          make conan-install-test
-
       - name: Build with coverage (Linux amd64 only)
         if: matrix.coverage == true
         run: |
+          . ./venv/bin/activate
+          GENERATORS="$(pwd)/build-conan/build/Release/generators"
```
SebastianMunozP: Before my changes, the CI on Linux was a pure native cmake build — no Conan involved at all. The Docker image had everything pre-installed as system packages (`librealsense`, `viam-cpp-sdk`, etc.), so cmake just called find_package and found them in the standard system paths (`/usr/lib`, `/usr/include`). No generators directory existed.

Now that CI runs `make conan-install-test` first, Conan downloads/resolves all dependencies and writes the generators directory. cmake then needs to be told "don't use system packages, use Conan's" — which is done by pointing it at the generators via CMAKE_TOOLCHAIN_FILE and CMAKE_PREFIX_PATH.

CMakeLists.txt
```diff
@@ -32,14 +32,7 @@ find_package(realsense2 REQUIRED)
...
 find_package(CURL REQUIRED)
-find_package(PkgConfig REQUIRED)
-
-# Use pkg-config for libzip when available (more reliable with system packages)
-# Fall back to find_package for Conan-provided libzip
-pkg_check_modules(LIBZIP IMPORTED_TARGET libzip)
-if(NOT LIBZIP_FOUND)
-    find_package(libzip REQUIRED)
-endif()
+find_package(libzip REQUIRED)
```
SebastianMunozP, line 35: Now that both CI and local development uses Conan to satisfy dependencies, we don't need the previous pkg_check_modules/conan fallback strategy, only Conan is sufficient now.

CMakeLists.txt
```diff
@@ -61,7 +54,7 @@ target_link_libraries(viamrealsense
         realsense2::realsense2
         libjpeg-turbo::turbojpeg-static
         CURL::libcurl
-        $<IF:$<TARGET_EXISTS:PkgConfig::LIBZIP>,PkgConfig::LIBZIP,libzip::zip>
+        libzip::zip
```
SebastianMunozP, line 57: Same here, we don't need to rely on PkgConfig no more for libzip, only Conan.

conanfile.py
```diff
@@ -21,7 +21,8 @@ class ViamRealsense(ConanFile):
     options = {"with_tests": [True, False]}
     default_options = {
         "with_tests": False,
-        "viam-cpp-sdk/*:shared": False
+        "viam-cpp-sdk/*:shared": False,
+        "libzip/*:shared": False
```
SebastianMunozP, line 25: Adding libzip as a statically linked dependency

.github/workflows/build-publish.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           make module.tar.gz
 
       - name: Upload module artifact
-        if: github.event_name == 'release'
+        if: github.event_name == 'release' || github.event_name == 'pull_request'
```
SebastianMunozP, line 81: Taking this change to upload our module.tar.gz to github on all PRs, not only on release only. This helps verify that the modules built by our CI is working OK. These artifacts are stored in GH for 90 days.

.github/workflows/test.yml
seanavery: Do we need to add a conan cache system here as well?

seanavery (commented):
Nice build refactor! Left one comment about caching conan builds for test.yml

.github/workflows/test.yml
SebastianMunozP: You mean to automatically store binaries to conan if we ever build from source in CI?

If that's the case, I'd like to tackle that in a separate PR

.github/workflows/test.yml
```diff
@@ -15,12 +15,12 @@ on:
 jobs:
   run-tests:
     name: Run Tests
-    timeout-minutes: 10
+    timeout-minutes: 30
```
seanavery: Is the conan build slower than native?

seanavery (approved):
LGTM

.github/workflows/test.yml
SebastianMunozP: Not at all, I bumped that while I was manually adding the missing precompiled binaries to our conan repo. I can revert it to 10.

PR #138, opened 2026-04-18
[RSDK-13785] Add udev rules and first_run hook for RealSense USB permissions

SebastianMunozP:
## Summary

- Adds `99-realsense-libusb.rules` (sourced from the librealsense repo) so RealSense camera device nodes are accessible to non-root users
- Adds `install_udev_rules.sh` to copy the rules to `/etc/udev/rules.d/` and reload udevadm
- Adds `first_run.sh` as the module's `first_run` hook — runs `install_udev_rules.sh` with sudo on Linux automatically on first start
- Updates `meta.json` to register `first_run.sh` as the `first_run` entrypoint
- Updates `CMakeLists.txt` and `conanfile.py` to include these files in the module bundle

## Why

On Linux, RealSense camera device nodes may not be accessible to non-root processes without udev rules. Since viam currently runs modules as root, this does not fix any known breakage — it is defensive hygiene that ensures tools like rs-enumerate-devices work without sudo, and future-proofs against any change to viam's execution model.

## What the rules file covers

`99-realsense-libusb.rules` handles three categories of devices:
- **USB devices** — all D400/L500/T200 series cameras, sets `MODE=0666` and `GROUP=plugdev` so any member of `plugdev` can open them
- **DFU/recovery devices** — cameras in firmware recovery mode get the same permissions so they can be reflashed
- **IIO/HID devices** — accelerometer and gyroscope interfaces on cameras with an IMU (e.g. D435i), including a kernel 4.15+ workaround that resets the accelerometer buffer enable flag on plug-in

## Test plan

- [x] Deploy to a fresh Linux install — confirm `first_run.sh` runs automatically and installs rules to `/etc/udev/rules.d/`
- [x] Verify `first_run.sh` and `99-realsense-libusb.rules` are included in `module.tar.gz`
- [x] Confirm no regressions in camera usage


first_run.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+OS=$(uname)
+
+if [[ "$OS" == 'Linux' ]]; then
+    SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
+    sudo "$SCRIPT_DIR/install_udev_rules.sh"
```
seanavery: Could this hang on password prompt?

SebastianMunozP: It shouldn't, as viam runs its modules as root. 

We have been doing this in Orbbec for a long time: https://github.com/viam-modules/orbbec/blob/main/first_run.sh#L11

seanavery: Ok sounds good. Not a common path, but, what happens if viam-server is ran as user?

SebastianMunozP: That's a good point, and unfortunately it is in that scenario where these rules are needed the most.

The problem is that we need to be root to install the rules, there's no way around that. But what I'll do is detect that we are not root in this script, error out and inform the user that it needs to manually install these rules, with clear instructions on how to do it.

first_run.sh
```diff
@@ -0,0 +1,29 @@
...
+
+if [[ "$OS" == 'Linux' ]]; then
+    SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
+    if [[ "$EUID" -eq 0 ]]; then
+        # Already root — run directly, no sudo needed
+        echo "Running as root, installing udev rules directly"
+        "$SCRIPT_DIR/install_udev_rules.sh"
+    else
+        # Non-root — attempt passwordless sudo
+        sudo -n "$SCRIPT_DIR/install_udev_rules.sh" 2>/dev/null
```
seanavery: should we capture error msg here and print out in the failure case?

`err=$(sudo -n "$SCRIPT_DIR/install_udev_rules.sh" 2>&1)`

SebastianMunozP: Good idea, just added that.

seanavery (approved):
LGTM

PR #139, opened 2026-04-18
[RSDK-13785] Bump librealsense to 2.57.7 for kernel 6.12 support

SebastianMunozP:
## Summary

- Bumps librealsense from `2.56.5` / `2.57.6` (platform-split) to a single unified `2.57.7` in `conanfile.py`
- Updates `Dockerfile.debian.bookworm` and `Dockerfile.cuda.debian.bookworm` to `v2.57.7`

## Why

librealsense 2.56.5 does not support Linux kernel 6.12, which ships with recent Raspberry Pi OS on the Pi 5. Attempting to open the camera produces a `bad optional access` crash inside librealsense. Version 2.57.7 supports kernel 6.12.

The previous `conanfile.py` used a platform split (2.57.6 on macOS, 2.56.5 elsewhere) — this removes the split and uses 2.57.7 everywhere. The Conan package for 2.57.7 is published to the `viamconan` JFrog registry for `linux/arm64`, `linux/amd64`, and `macOS`.

## Notes

- librealsense 2.57.7 is labeled "Beta" by Intel but is focused on stability; it is currently the only version with kernel 6.12 support
- As part of this work I also added a new Conan recipe for version 2.57.7,  where I had to includes patches for GCC 12/13 missing includes (`<cstring>`, `<algorithm>`, `<stdexcept>`)
- I was a good samaritan and [created a PR](https://github.com/realsenseai/librealsense/pull/14944) for those includes in librealsense upstream, which was accepted and merged

## Test plan

- [x] Build `module.tar.gz` for `linux/arm64` — confirm successful build with librealsense 2.57.7
- [x] Confirm camera works on Raspberry Pi 5 (kernel 6.12) — no `bad optional access` crash
- [x] Confirm camera works on `linux/amd64`

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -37,10 +37,7 @@ class ViamRealsense(ConanFile):
 
     def requirements(self):
         self.requires("viam-cpp-sdk/0.31.0")
-        if self.settings.os == "Macos":
-            self.requires("librealsense/2.57.6")
-        else:
-            self.requires("librealsense/2.56.5")
+        self.requires("librealsense/2.57.7")
         self.requires("libjpeg-turbo/[>=2.1.0 <3]")
         self.requires("libcurl/[>=8.0.0 <9]")
         self.requires("libzip/1.11.1")
--- a/etc/Dockerfile.cuda.debian.bookworm
+++ b/etc/Dockerfile.cuda.debian.bookworm
@@ -84,7 +84,7 @@ RUN apt install -y libgtest-dev
 RUN mkdir -p /root/opt/src
 
 # install librealsense from source
-ENV REALSENSE_VERSION="v2.56.5"
+ENV REALSENSE_VERSION="v2.57.7"
 RUN cd /root/opt/src && \
     git clone https://github.com/IntelRealSense/librealsense.git && \
     cd librealsense && \
--- a/etc/Dockerfile.debian.bookworm
+++ b/etc/Dockerfile.debian.bookworm
@@ -79,7 +79,7 @@ RUN apt install -y libgtest-dev
 RUN mkdir -p /root/opt/src
 
 # Install librealsense from source
-ENV REALSENSE_VERSION="v2.56.5"
+ENV REALSENSE_VERSION="v2.57.7"
 RUN cd /root/opt/src && \
     git clone https://github.com/IntelRealSense/librealsense.git && \
     cd librealsense && \


conanfile.py
```diff
@@ -37,10 +37,7 @@ def validate(self):
 
     def requirements(self):
         self.requires("viam-cpp-sdk/0.31.0")
-        if self.settings.os == "Macos":
-            self.requires("librealsense/2.57.6")
-        else:
-            self.requires("librealsense/2.56.5")
+        self.requires("librealsense/2.57.7")
         self.requires("libjpeg-turbo/[>=2.1.0 <3]")
```
hexbabe, line 41: this is new

why needed?

SebastianMunozP: It comes from this [thread](https://viaminc.slack.com/archives/C059QREJ6DR/p1776452453996949), and its corresponding [ticket](https://viam.atlassian.net/browse/RSDK-13785?atlOrigin=eyJpIjoiZmFmNmYwODU5MGFkNGMxY2I2OTY0MTFmNWMyNmYwMGMiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

There's a customer trying to use a Realsense D435i in a Pi5 but couldn't, the module could not communicate with the camera. I did some work over the weekend, and found out that from May last year, the Linux Kernel of Bookworm Pi5 images is 6.12, same thing starting from October last year for Trixie.

The Realsense stable release we are using ([2.56.5](https://github.com/realsenseai/librealsense/releases/tag/v2.56.5)) supports up to kernel 6.8, which causes it not to work with this camera.

This [latest beta release](https://github.com/realsenseai/librealsense/releases/tag/v2.57.7) supports that version and from its release notes: "2.57.7 release was focused on stability and quality"

conanfile.py
```diff
@@ -37,10 +37,7 @@ def validate(self):
 
     def requirements(self):
         self.requires("viam-cpp-sdk/0.31.0")
-        if self.settings.os == "Macos":
```
seanavery, line 41: [question] why did we have a diff realsense sdk version for macos beforehand?

seanavery (approved):
LGTM. Nice repro & fix!

conanfile.py
hexbabe: thanks for the context

i was specifically asking about why direct dep libjpeg is needed

SebastianMunozP: realsense delivers its RGB frames as raw data, so we need a library to compress it into JPEG.

conanfile.py
SebastianMunozP: Good question,

The support of Mac on realsense is not realsense's priority, so they list this as supported: [Mac OS (High Sierra 10.13.2 & 12+)(4)](https://github.com/realsenseai/librealsense/releases), but with the note that: "(4) This is compilable, but not validated"

And in reality, the released versions before 2.56.6  did not work for the D435 camera.

Now, a little bit of history: To overcome this, in late December/ early January this year I created a special librealsense in our viamconan called (librealsense+viam) that consisted on a base image + patches to support Mac, that worked well but with the caveat that it was not an official release from realsense.

Then, realsense released 2.56.6 that contained the patches I manually applied, so I moved Mac to that version.

Since 2.56.6 is a beta release, I preferred to consume it conditionally on Mac only, as I didn't have a compelling reason (until now) to bump the other platforms to a beta release instead of staying in our current stable.

conanfile.py
hexbabe: I am confused about how we did this before if we didn't include this dependency here

hexbabe: this seems orthogonal to resolving the customer issue

hexbabe (approved):
looks good to me, but I have some questions

conanfile.py
SebastianMunozP: I'm not touching libjpeg in this PR if that's what you are asking about, I got kind of confused.

hexbabe: sorry maybe I'm colorblind I misread the diff

conanfile.py
seanavery: Got it, thanks for the context

PR #140, opened 2026-04-19
[RSDK-13787] Fix camera not resuming after firmware is already up to date

SebastianMunozP:
## Summary:

When do_command({"update_firmware": ""}) is called, it excercises the auto-detect flow in which we detect the current firmware of the device and its recommended one and if they differ and we have stored the URL of the recommended firmware, we download it and install it.  

The current issue is that if they are the same, i.e. the camera is already at its recommended firmware, we stop its streaming anyways and don't recover unless the camera is re-plugged.

The dis consists on that In auto-detect mode we now check the current and recommended firmware versions before touching any device state. If they match, return a success response immediately instead of stopping the streaming.

Also, adding unit tests to cover update_firmware

## Test plan:

- [x] Call do_command({"update_firmware": ""}) on a camera already running the recommended firmware version
- [x] Verify the command returns successfully with the "already at recommended version" message
- [x] Verify the camera resumes streaming immediately after without requiring a module restart

### DIFF
--- a/src/module/realsense.hpp
+++ b/src/module/realsense.hpp
@@ -1004,8 +1004,8 @@ private:
           return response;
         }
 
-        // If no firmware URL was provided, check whether the device already runs
-        // the recommended version before stopping the stream
+        // If no firmware URL was provided, check whether the device already
+        // runs the recommended version before stopping the stream
         if (firmware_url.empty()) {
           auto device_guard = device_->synchronize();
           auto pre_check_device = device_guard->device;
--- a/src/module/realsense.hpp
+++ b/src/module/realsense.hpp
@@ -1037,18 +1037,29 @@ private:
 
       // Check if firmware update succeeded
       if (update_result.first) {
-        // Success - clear device assignment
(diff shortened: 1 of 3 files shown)


src/module/realsense.hpp
```diff
@@ -182,6 +182,9 @@ struct DeviceFunctions {
       std::shared_ptr<boost::synchronized_value<device::ViamRSDevice<>>> &,
       realsense::RsResourceConfig const &, viam::sdk::LogSource &)>
       reconfigureDevice;
+  std::function<bool(std::shared_ptr<rs2::device>, std::string &current,
```
seanavery, line 185: [question] This is for mockability in tests?

seanavery (approved):
LGTM

Seems much cleaner to have the prefilight check and noop handling moved to realsense.hpp

src/module/realsense.hpp
SebastianMunozP: That's correct.

PR #141, opened 2026-04-29
fix(get_properties): anchor extrinsics to depth frame; update README

SebastianMunozP:
**This fix addresses an edge case that affects only users who explicitly configure `sensors: ["depth", "color"]` or `sensors: ["depth"]`. The default sensor list is `["color", "depth"]`, which was already producing correct extrinsics — so the vast majority of users are unaffected.**

### Bug fix: `get_properties` extrinsics

`get_properties` was using the "main sensor" (first entry in the `sensors` list) to decide which stream to anchor extrinsics from. This caused two problems:

- **`sensors: ["depth", "color"]`** — returned `depth.get_extrinsics_to(color)` ≈ `{+14.7, 0, 0}` mm instead of zero, because depth was treated as the reference frame when it should always be the anchor
- **`sensors: ["depth"]`** — crashed with an exception because `profile.get_stream(RS2_STREAM_COLOR)` throws when color is not active

The fix:
- Wraps both `get_stream()` calls in try-catch so single-sensor configs no longer crash
- Always prefers color intrinsics when color is configured, regardless of sensor order
- When color is present, extrinsics are `color → depth_left_imager` (the camera reference frame), as expected
- When only depth is configured, extrinsics are identity (depth IS the reference frame)
- Removes `getMainSensor()`, which was the source of the ordering-dependent behavior

| Config | Before | After |
|---|---|---|
| `["color", "depth"]` | color intrinsics, correct extrinsics ✓ | unchanged ✓ |
| `["depth", "color"]` | depth intrinsics, **wrong extrinsics (+14.7 mm)** | color intrinsics, zero extrinsics ✓ |
| `["depth"]` | **crash** | depth intrinsics, zero extrinsics ✓ |

### Why this matters: consistency with `get_geometries`

`get_geometries` unconditionally anchors bounding-box geometry to the **depth left imager**, regardless of sensor config. Before this fix, `get_properties` could return extrinsics anchored to a different origin depending on sensor order — making the two APIs inconsistent with each other for affected configs. Any service combining spatial information from both APIs (e.g. a vision service composing color-based detection poses against world-frame transforms derived from `get_geometries`) would silently produce wrong 3D positions, with the ~14.7 mm X offset propagating into arm targeting, obstacle avoidance, or any spatial reasoning that mixes visual detection with depth-based geometry.

After this fix, both APIs agree on the same reference frame (depth left imager) for all sensor configurations.

### README

- Corrected the `sensors` attribute description: removed the "main sensor" framing and the `get_image` reference (that method is deprecated and not implemented)
- Added a new **"Frame origin and `extrinsic_parameters`"** section explaining:
  - The camera reference frame is always the depth left imager, consistent with `get_geometries`
  - `get_properties` always returns color intrinsics when color is configured
  - How to convert a color-frame detector pose (AprilTag, PnP) into the camera reference frame using `extrinsic_parameters.translation`
  - The ~15 mm X baseline on D435/D435i and why skipping this correction produces a visible offset


src/module/realsense.hpp
```diff
@@ -714,10 +711,20 @@ class Realsense final : public viam::sdk::Camera,
...
-        auto color_stream = profile.get_stream(RS2_STREAM_COLOR)
-                                .as<rs2::video_stream_profile>();
+        // Fetch streams defensively: a stream absent from the active profile
+        // (e.g. sensors: ["depth"] has no color stream) throws rather than
+        // returning an invalid profile, so we catch and leave unset.
+        rs2::video_stream_profile depth_stream, color_stream;
+        try {
+          depth_stream = profile.get_stream(RS2_STREAM_DEPTH)
+                             .as<rs2::video_stream_profile>();
+        } catch (...) {
```
erh: should we be logging?

btshrewsbury-viam: Added debug sysout, may need to be pushed to error if we run into issues

erh (approved):

README.md
```diff
@@ -82,6 +82,36 @@ The following methods of the Viam camera API are supported:
...
+
+# pose_t is the detector's translation output in meters, in the color frame.
+pose_in_camera_frame_mm = (
+    pose_t[0] * 1000 + ox,
+    pose_t[1] * 1000 + oy,
+    pose_t[2] * 1000 + oz,
+)
+```
+
+Skipping this step produces a offset in X equal to the color-to-depth baseline (roughly 15 mm on D435/D435i), visible as soon as the camera frame is composed against the world frame or other components.
```
seanavery: spelling: a->an

README.md
```diff
@@ -43,7 +43,7 @@ The following attributes are available for `viam:camera:realsense` cameras:
 
 | Name | Type | Inclusion | Description |
 | ---- | ---- | --------- | ----------- |
-| `sensors` | list | Optional | The RealSense data streams you want your robot to sense from. A list that contain the strings `color` and/or `depth`. The sensor that comes first in the list is designated the "main sensor", and is the image that gets returned by `get_image` calls and appears in the **CONTROL** tab on the [Viam app](https://app.viam.com). If you would like a list of images from all listed sensors simultaneously, use [`GetImages`](https://docs.viam.com/components/camera/#getimages).  |
+| `sensors` | list | Optional | The RealSense data streams you want your robot to sense from. A list containing the strings `color` and/or `depth`. The sensor listed first determines the intrinsics and extrinsics direction returned by `get_properties` (e.g. `["color", "depth"]` gives color intrinsics with color→depth extrinsics, while `["depth", "color"]` gives depth intrinsics with depth→color extrinsics). Use [`GetImages`](https://docs.viam.com/components/camera/#getimages) to retrieve images from all listed sensors simultaneously. Defaults to `["color", "depth"]` if omitted. |
```
seanavery: Looks like this needs to be updated for the new get_properties contract, where the camera frame is always the depth sensor

src/module/realsense.hpp
seanavery: This needs to be changed to fulfill the contract of always using depth sensor as the reference frame right?
(identity return)

```cpp
          auto ref_stream = depth_stream;
```

seanavery (commented):
Crash protection change looks good.  But I don't think the get_properties logic actually enforces the new "always anchor to depth" contract.

10zingpd:
@seanavery any idea if this pr is still good? if so can you help close this?

seanavery:
@10zingpd Thanks for bumping, I think this PR needs a little sprucing up before merging

PR #142, opened 2026-05-07
Add align_color_depth attribute for get_images

bhaney:
Add opt-in align_color_depth config to align depth to color in get_images

GetImages currently returns the color and depth streams straight from the device's frameset, with no alignment between them. The IR/depth and RGB sensors on a RealSense are physically offset (~15-55 mm depending on model) and have different fields of view, so a 2D mask drawn on the color image cannot be used to look up depth values via depth_image[v, u] -- the same pixel coordinates refer to different physical points in the two streams. Consumers that combine color-image segmentation with depth (e.g. SAM2-based pipelines) end up reading depth from the wrong pixel and producing distorted point clouds.

Add a new optional bool config attribute align_color_depth (default false) that, when true, runs librealsense's rs2::align(RS2_STREAM_COLOR) on the frameset inside get_images() so the depth array is resampled to the color frame. Default is left at false to preserve existing behavior for production deployments.

GetPointCloud is unchanged: it already deprojects depth and color-maps via rs2::pointcloud and is correct as-is.


HipsterBrown (approved):

HipsterBrown:
Probably need to see why the MacOS test is failing

PR #147, opened 2026-05-26
expose config parameters for the IR sensor of the realsense

npmenard:


src/module/realsense.hpp
```diff
@@ -898,6 +1094,58 @@ class Realsense final : public viam::sdk::Camera,
       }
     }
 
+    if (attrs.count("laser_power")) {
+      if (not attrs["laser_power"].is_a<double>()) {
+        throw std::invalid_argument("laser_power must be a number (0..360)");
+      }
+      double lp = attrs["laser_power"].get_unchecked<double>();
```
seanavery, line 1157: should we apply validation to the DoCommand path as well?

seanavery (approved):
Nice!

Have we manually tested both config and DoCommand path? Assuming it does not matter if the pipeline is active or not given the official RealSense Viewer has a live tuning feature.

Also might be worth clauding some sanity unit tests for DoCommands and config validation

src/module/realsense.hpp
npmenard: we bubble up the error from rs2 lib

PR #150, opened 2026-06-09
ci(test): share Conan cache with build-publish and raise timeout

ale7714:
## Summary

`test.yml` had **no Conan caching**, so the `macos-latest` leg rebuilt all heavy deps (grpc, protobuf, openssl, re2) from source on every run. That pushed it past the 10-minute `timeout-minutes`, and with the matrix default `fail-fast`, the timeout cancelled the sibling Linux legs too — masking their actual pass/fail (this is what cancelled the recent `main` runs).

## Changes

- **Reuse `build-publish.yml`'s Conan cache.** Add a `Restore cached libraries` step with the *identical* `~/.conan2` path and `${{ matrix.os }}-conan-${{ hashFiles('**/conanfile.py') }}` key. GitHub matches caches by key across workflows in the same repo, so:
  - `macos-latest` collides between both workflows → the macOS cache is genuinely **shared** (no separate cache created).
  - The Linux legs use different os names / container images, so they keep their own cache namespace — which is correct (don't mix Conan binaries across base images).
  - `restore-keys` gives partial hits when `conanfile.py` changes, so only the changed dep rebuilds.
- **Raise `timeout-minutes` 10 → 30.** A cold macOS build must complete once to actually *save* the cache; at 10m it was cancelled before the save, so the cache could never warm up.

`fail-fast` is intentionally left at the default per review feedback.

## Notes

- Uses `actions/cache@v3` to match `build-publish.yml`'s existing style. SHA-pinning all actions is a separate repo-wide cleanup.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   run-tests:
     name: Run Tests
-    timeout-minutes: 10
+    timeout-minutes: 30
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -38,6 +38,16 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+
+      # Same path/key as build-publish.yml so the two workflows share Conan
+      # caches (macos-latest collides; Linux legs keep their own namespace).
+      - name: Restore cached libraries
+        uses: actions/cache@v3
+        with:
+          path: ~/.conan2
+          key: ${{ matrix.os }}-conan-${{ hashFiles('**/conanfile.py') }}
+          restore-keys: |
+            ${{ matrix.os }}-conan-
 
       - name: Install dependencies
         run: |


.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ on:
 jobs:
   run-tests:
     name: Run Tests
-    timeout-minutes: 10
+    timeout-minutes: 30
```
ale7714, line 18: mac os seem to be really slow suddenly and it's hard to debug because the job gets cancelled before even finishes.

seanavery (approved):
Cache and timeout bump LGTM.

But curious why we are not able to pull the darwin build without rebuilding deps...

[cpp-sdk darwin conan publish](https://github.com/viamrobotics/viam-cpp-sdk/blob/f3552c085227bd53e1af07ecfb6598ee4fa4039c/.github/workflows/conan-publish.yml#L23)

ale7714:
i think the building is succeeding but test timeout

PR #151, opened 2026-06-09
fix(conan): pin drifting deps to prebuilt versions and add lockfile

ale7714:
## Problem

CI has been timing out since **Jun 8** — every platform (incl. the prebuilt Linux containers) spends ~30 min compiling grpc/protobuf/openssl/re2 **from source** in `Install Conan dependencies`, then gets cancelled before the test step.

**The repo didn't change.** Between the last green build (#146, May 13) and the first failure (#141, Jun 8) there was one source/README commit and an **empty `conanfile.py` diff**. The regression is entirely upstream:

- The SDK pulls grpc/protobuf/re2/abseil/openssl/zlib via **floating version ranges** (`grpc/[>=1.48.4]`, etc.) and the repo has **no lockfile**, so every build re-resolves against whatever is newest in the remotes.
- After May, newer upstream releases published (grpc `1.72→1.78.1`, re2 `→20251105`, protobuf `→5.29.6/6.33.5`, …). `viamconan` has the **old** versions prebuilt, not the new ones → `--build=missing` recompiles from source.

Same code, different result over time — the build wasn't reproducible.

## Fix

1. **Pin the drifting deps** in `conanfile.py` to the versions from the May 13 green build (all prebuilt in `viamconan`): grpc 1.72.0, protobuf 5.27.0, abseil 20250127.0, re2 20230301, openssl 3.6.0, zlib 1.3.1 (xtensor already pinned). Conan `Download`s these instead of `Build`ing them.
2. **Commit `conan.lock`** freezing the full resolved graph (versions + recipe revisions), generated against `viamconan` with CI's settings (`compiler.cppstd=17`, `build_type=Release`). This pins the long-tail transitives too (boost, libcurl, …) so a future upstream publish can't silently drift the build again. Conan 2 auto-consumes `conan.lock` from the repo root.

Bonus: pinning collapses the **dual protobuf** (5.29.6 + 6.33.5) back to a single 5.27.0 in both host and build context — visible in the lockfile's `build_requires`.

## Validation

The lockfile was generated by resolving against `viamconan`; the heavy deps resolved to the exact green-build recipe revisions (`grpc/1.72.0#716982a6…`, `openssl/3.6.0#89e8af1d…`, etc.). CI on this PR should now `Download` the deps and finish in minutes instead of timing out.

## Note

The durable owner of this is the SDK (its floating ranges are the root cause). This pins at the module level to unblock CI now; longer term the SDK should bound its grpc/protobuf ranges or ship its own lock.


ale7714:
i'm going to merge to unblock ci, i'll follow up with cloud build

PR #155, opened 2026-06-29
Add distortion params to get_properties()

bgrigsby8:
Adds outputting distortion params to the realsense module

It seems like this was blocked until https://viam.atlassian.net/browse/RSDK-12408 was done. Now that it is done, the realsense module can now export how it reports its distortion parameters, which is in inverse brown conrady format. 

Any use of these distortion parameters should then be handled by whoever is calling get_properties.

The other thing is that the inverse brown conrady distortion parameters reported by realsense are reported in {k1, k2, p1, p2, k3} order, but rdk put it in {k1, k2, k3, p1, p2} format, so a reordering was needed to output in Viam.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ build
 venv
 .venv
 .VIAM_RELOAD_ARCHIVE.tar.gz
+compile_commands.json
+.cache
+CMakeUserPresets.json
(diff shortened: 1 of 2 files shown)


seanavery (approved):
nice, LGTM

PR #158, opened 2026-06-30
[RSDK-13964] - Add rs2 conan publish

seanavery:
## Description

- Adds librealsense and libusb conan upload workflow

## Tests

- cold `viam module reload` went from `~55min` to `<10min`
```
 …  Preparing for build...
 ✓     → Source code archive created (0s)                                                     
 ✓     → Source code uploaded (2s)                                                            
 ✓  Prepared for build (2s)
 …  Building...
 ✓     → Build started (ID: 60b42527-3fdf-4860-bb2e-785e34b3463e) (3s)                        
 ✓     → Spin up environment (40s)                                                            
 ✓     → Install Viam CLI (2s)                                                                
 ✓     → Install dependencies (2s)                                                            
 ✓     → Download repository archive (2s)                                                     
 ✓     → Build module (8m3s)                                                                  
 ✓     → Uploading artifacts (10s)                                                            
 ✓  Built (9m3s)
 …  Reloading to part...
 ✓     → Module added to part (0s)                                                            
 ✓  Reloaded to part (0s)
```

- warm reloads `~45min` -> `<3min`

```
 …  Preparing for build...
 ✓     → Source code archive created (0s)                                                     
 ✓     → Source code uploaded (2s)                                                            
 ✓  Prepared for build (2s)
 …  Building...
 ✓     → Build started (ID: 5831ab03-b528-4495-895b-d1ffc808751f) (3s)                        
 ✓     → Spin up environment (52s)                                                            
 ✓     → Install Viam CLI (2s)                                                                
 ✓     → Install dependencies (4s)                                                            
 ✓     → Restoring cache (2s)                                                                 
 ✓     → Build module (1m48s)                                                                 
 ✓     → Run model detection (2s)                                                             
 ✓     → Uploading artifacts (8s)                                                             
 ✓  Built (3m0s)
 …  Reloading to part...
 ✓     → Module added to part (0s)                                                            
 ✓  Reloaded to part (0s)
 ```

- manually verified build works on amd64 machine

### CODE REVIEW

File: conanfile.py [hunk 1]
```diff
@@ -44,6 +44,7 @@ def requirements(self):
         self.requires("libzip/1.11.1")
         # Pin to prebuilt versions; the SDK's floating ranges drifted to
         # newer releases that aren't prebuilt, forcing slow source builds.
+        self.requires("boost/1.90.0", override=True)
```
seanavery: viam-cpp-sdk's floating boost range resolves to 1.91.0, which has no prebuilt binary on viamconan, so boost builds from source, and since it's an SDK dep the SDK rebuilds too. Worked around it by pinning boost/1.90.0

.github/workflows/conan-publish.yml
```diff
@@ -0,0 +1,75 @@
...
+            --lockfile conan.lock \
+            --build=missing
+
+      - name: Upload librealsense and libusb binaries
+        run: |
+          . ./venv/bin/activate
+          conan upload "librealsense/*" -r viamconan -c
+          conan upload "libusb/*" -r viamconan -c
+
+      - name: Upload module recipe + binary (optional)
```
seanavery: Remove this

.github/workflows/conan-publish.yml
```diff
@@ -0,0 +1,75 @@
...
+          DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+            make sudo git ca-certificates
+
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Run setup (installs conan + adds viamconan remote)
+        run: make setup
+
+      - name: Verify viamconan auth (fail fast before the long build)
```
seanavery, line 44: Should we remove this fail fast auth?

lia-viam: I think fail fast is good here, no sense firing off a long build which then won't be able to upload

conan.lock
```diff
@@ -25,6 +25,7 @@
         "c-ares/1.34.6#545240bb1c40e2cacd4362d6b8967650%1770059815.568",
         "bzip2/1.0.8#00b4a4658791c1f06914e087f0e792f5%1768277099.91",
         "boost/1.91.0#ea540ca2133d831b560036aa24dece3c%1779143123.044",
+        "boost/1.90.0#d5e8defe7355494953be18524a7f135b",
```
lia-viam: having two copies of boost is a bit odd, maybe investigate this or point claude to it? note also this only has a `#revision` but not a `%timestamp` which i think can happen when you first add a dependency to a lockfile but haven't yet done a build with it.

conanfile.py
```diff
@@ -44,6 +44,7 @@ def requirements(self):
         self.requires("libzip/1.11.1")
         # Pin to prebuilt versions; the SDK's floating ranges drifted to
```
lia-viam, line 45: this is not wrong strictly speaking, and i see that some of these `override=True` lines are not part of the diff for this PR specifically, but i think `override=True` is not the best way to solve this and it would be better to use the lockfile. the conanfile.py should specify how a package can be built, potentially for different use cases or consumers, and the lockfile should be used for CI/build reproducibility purposes. 

i think try removing the `override=True` pinned requirements but build with the `conan.lock` and see if that doesn't still give you the fast build times. it looks like the lockfile is indeed pinned to the versions currently listed below.

lia-viam (changes requested):

conan.lock
seanavery: Yeah I had some boost rebuild issue at some point and ended up in this bad state. 

Removed dup and pinning in lockfile now - cache hit during build works just fine.

conanfile.py
seanavery: Removed conanfile.py diff - just using same lock from main branch and everything is working fine

conan.lock
```diff
@@ -24,7 +24,7 @@
         "grpc/1.72.0#716982a6240893b25ccedd8de6108462%1769627228.784",
         "c-ares/1.34.6#545240bb1c40e2cacd4362d6b8967650%1770059815.568",
         "bzip2/1.0.8#00b4a4658791c1f06914e087f0e792f5%1768277099.91",
-        "boost/1.91.0#ea540ca2133d831b560036aa24dece3c%1779143123.044",
```
seanavery, line 27: Lock still has boost downgrade... perhaps i should revert to the upstream 1.91.0 version?

seanavery: Looks like check is >=1.74.0 so we should 1.90.0 use should be good
https://github.com/viamrobotics/viam-cpp-sdk/blob/4f610d684cb8e200160c06deba6d3eb693ee439d/conanfile.py#L76-L82

seanavery: @lia-viam I think the 1.91.0 version actually does incur a rebuild. Do we need to match the exact version bundled in the cpp-sdk conan upload?

lia-viam: A rebuild on the first time is expected, after that it should be in the cache once you've uploaded! And yes I would recommend deleting 1.90.0 and having the 1.91.0 entry match that in the SDK lockfile.

seanavery: Ok, 1.90.0 is actually the correct match-up to the cpp-sdk repo:
`boost/1.90.0#d5e8defe7355494953be18524a7f135b%1765955095.179`
https://github.com/viamrobotics/viam-cpp-sdk/blob/6c4ad7c07776cc45c3d5b92b3a0cf92f28ec8cf6/etc/conan/conan.lock#L19

lia-viam (approved):

PR #159, opened 2026-07-07
[RSDK-13700] - Remove reconfigure

seanavery:
## Description

- Large bump in the cpp-sdk from 0.31.0 -> 0.38.1
- Removing reconfigure and related helpers

Hopefully the auto-bump CI runner will work well with smaller sdk diffs: https://github.com/viam-modules/viam-camera-realsense/pull/154


src/module/realsense.hpp
```diff
@@ -445,6 +340,18 @@ class Realsense final : public viam::sdk::Camera,
...
+    viam::sdk::ProtoStruct status;
+    status["physical_camera_assigned"] = physical_camera_assigned_.get();
+    status["is_recovery_mode"] = is_recovery_mode_.get();
+    if (device_) {
+      auto device_guard = device_->synchronize();
+      status["serial_number"] = device_guard->serial_number;
+      status["streaming"] = device_guard->started;
+    }
+    return status;
+  }
```
stuqdog, line 353: This is out of scope for this pr but I'm surprised to see that all these methods are defined in header files. Curious if this is known/for a specific reason? cc @lia-viam

stuqdog (approved):
Looks reasonable to me!

src/module/realsense.hpp
seanavery: I believe it is setup this way for mock testing
(it's templated on the context type so tests can inject a fake rs2 context)


We already split some out into `_impl` headers (device.hpp & device_impl.hpp) so it's not all inline, but that's still header based

@lia-viam @stuqdog  Let me know if there is a better pattern we should be following - i can file a follow-up ticket

PR #160, opened 2026-07-10
Add D405 camera support

shannonbradshaw:
## What this does

Adds the D405 to the supported camera models. The D405 differs from the other supported D400 cameras in two ways the module previously could not handle, both verified against the pinned librealsense (2.57.7) source:

**1. No dedicated RGB sensor.** The D405's single stereo sensor serves both the color and depth streams, and it does not cast to `rs2::color_sensor` (librealsense sets `_separate_color = false` and never creates a color sensor object). Sensor discovery in `createSingleSensorConfig` / `createSwD2CAlignConfig` previously looked for one sensor casting to `rs2::color_sensor` and one to `rs2::depth_sensor`, which would leave a default-constructed sensor and fail. Discovery now selects streams by profile format (RGB8 / Z16) across all sensors, which behaves identically on D415/D435/D435i. Auto-exposure priority is disabled on whichever sensor serves the color stream (gating moved from the helper into the call sites). Non-video profiles (IMU motion streams on D435i) are skipped with an explicit `is<video_stream_profile>()` guard so they can never be mis-cast.

**2. 0.1 mm default depth units.** Every other D400 defaults to 1 mm depth units; the D405 defaults to 100 µm (`d400-device.cpp`, PID-gated). Raw Z16 values were previously copied straight into `image/vnd.viam.dep`, which encodes millimeters, so D405 depth would have read 10x too far. `encodeDepthFrameToResponse` now rescales using the frame's reported depth units, with a pass-through fast path for 1 mm devices and saturation at the uint16 max. `GetPointCloud` needed no change (`rs2::pointcloud` applies the depth scale internally).

Also included:

- `get_geometries` entry for the D405: pose `{9, 0, -7.8}`, box `{42, 42, 23}` mm, from datasheet 337029-017 (Tables 3-54, 4-16, 4-20). The origin is the left imager itself since the D405's color stream comes from it, so the color->depth baseline is ~0. The imagers sit on the housing's horizontal centerline (verified by measuring the 1:1 mechanical drawing in Figure 4-8, cross-checked against the 18 mm stereo baseline).
- README notes on D405 frame semantics (shared color/depth origin) and the supported-hardware table.
- Firmware auto-detect needs no change: since the 5.13 unification the D400-series firmware line covers the D405, and librealsense 2.57.7 reports 5.17.0.10 as recommended for all D400 PIDs.

## Testing

- Full unit suite passes (146/146), including 9 new tests: D405 single-sensor topology (dual-stream and color-only configs), graceful nullptr when no color stream exists, depth-scale rescaling with exact encoded bytes, saturation, 1 mm pass-through equivalence, invalid-scale rejection, and D405 model-string parsing.
- **Not yet validated on a physical D405.** Streaming end-to-end, the depth-units behavior on real firmware, and actual reported extrinsics still need bench time with hardware; the README marks the D405 column as implemented but unvalidated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,14 +87,14 @@ The following methods of the Viam camera API are supported:
 
 The RealSense D4xx series has multiple imagers at slightly different positions on the device. The **camera frame origin is anchored at the depth left imager** for `extrinsic_parameters`.
 
-> **Note:** the bounding-box geometries returned by `get_geometries` are anchored at the **color (RGB) sensor**, not the depth left imager, since `get_properties` reports color intrinsics and the poses callers derive from the images are in the color frame. This differs from the `extrinsic_parameters` reference frame described below, and from the `GetPointCloud` output (which is in the depth frame). If you compose geometries and point clouds in the same frame, account for the ~14.7 mm (D435/D435i) color→depth baseline.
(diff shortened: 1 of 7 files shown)


shannonbradshaw:
Closing in favor of #165, which is this same commit rebased onto current main and validated on a physical D405.

PR #161, opened 2026-07-10
fix(pointcloud): register point cloud to the color frame

HipsterBrown:
## Summary

`get_point_cloud()` built the cloud with `rs2::pointcloud::calculate(depth)` on an **unaligned** frameset, so the XYZ vertices came out in the **depth / left-IR sensor frame** — while `get_properties()` advertises **color** intrinsics. Consumers that project cloud points through those intrinsics *without* applying extrinsics (notably the `detections-to-segments` vision service, which explicitly documents that it assumes a color-registered cloud) therefore saw a depth↔color baseline shift. The result: 2D detection bounding boxes selected the **wrong** 3D points — a horizontal, depth-dependent misalignment (~30 px at 0.5 m on a D435).

The fix aligns depth to color with `rs2::align(RS2_STREAM_COLOR)` in `PointCloudFilter::process` before `calculate()`, so vertices are emitted in the color frame, consistent with the reported intrinsics. Stream-presence checks run first to preserve the existing "requires both color and depth" diagnostic. The change is unconditional (not gated behind `align_color_depth`, which only affects `get_images`): a depth-frame cloud paired with color intrinsics is internally inconsistent for any consumer.

- `src/module/device.hpp` — `PointCloudFilter` now holds an `rs2::align(RS2_STREAM_COLOR)` and registers depth→color before deprojecting.

## Test Plan

- [x] `make test` — 139/139 unit tests pass, module compiles.
- [x] **Hardware validation** on a live D435I (cam-1) via cloud `module reload`, using a reprojection-consistency probe (each point's texture-mapped RGB is ground truth for its true color pixel; the best-matching pixel shift reveals any frame offset):

  | metric (Z≈0.4–0.6 m, ~40k pts) | before | after |
  |---|---|---|
  | best horizontal shift `dx` | **29 px** | **0 px** |
  | color err @ 0 shift | 3.19 | 1.35 |
  | color err @ best shift | 1.59 | 1.35 |

  Predicted shift for a 15 mm baseline at 0.5 m was 27.2 px, matching the measured 29 px pre-fix. Post-fix, zero shift is optimal (`err@0 == err@best`), confirming the cloud is now color-registered.

## Notes

- Aligning depth→color trims the cloud to the depth∩color FOV and may leave small holes where depth is invalid — expected and correct for a color-registered cloud.

### DIFF
--- a/src/module/device.hpp
+++ b/src/module/device.hpp
@@ -22,6 +22,12 @@ public:
       : pointcloud_(std::make_shared<rs2::pointcloud>()),
         align_to_color_(std::make_shared<rs2::align>(RS2_STREAM_COLOR)) {}
   std::pair<rs2::points, rs2::video_frame> process(rs2::frameset frameset) {
+    // This body only runs against a live RealSense: every statement operates on
+    // librealsense frame/align/pointcloud objects that cannot be constructed or
+    // driven without hardware, so it is unreachable in the (camera-less) CI test
+    // suite and is excluded from coverage. Validated on-device via the alignment
+    // probe instead.
+    // LCOV_EXCL_START
     // Validate both streams are present first so we can surface a helpful
     // message (align_to_color_->process below would otherwise throw a generic
     // error when the color stream is missing).
@@ -54,6 +60,7 @@ public:
     pointcloud_->map_to(color_frame);
     auto points = pointcloud_->calculate(depth_frame);
     return std::make_pair(points, color_frame);
+    // LCOV_EXCL_STOP
   }
 
 private:
--- a/src/module/device.hpp
+++ b/src/module/device.hpp
@@ -18,20 +18,39 @@ namespace realsense {
 namespace device {
 class PointCloudFilter {
 public:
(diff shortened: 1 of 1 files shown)


src/module/device.hpp
```diff
@@ -18,27 +18,53 @@ namespace realsense {
...
-    if (!depth_frame) {
+    // This body only runs against a live RealSense: every statement operates on
+    // librealsense frame/align/pointcloud objects that cannot be constructed or
+    // driven without hardware, so it is unreachable in the (camera-less) CI
+    // test suite and is excluded from coverage. Validated on-device via the
+    // alignment probe instead. LCOV_EXCL_START Validate both streams are
+    // present first so we can surface a helpful message
+    // (align_to_color_->process below would otherwise throw a generic error
+    // when the color stream is missing).
+    if (!frameset.get_depth_frame()) {
```
seanavery, line 33: Might be a little defensive here given we already check depth frame exists in get_point_cloud before process is hit.

seanavery (approved):
Nice! 

Probably out of scope for this PR, but noticing that PointCloudFilter is not templated and testable like the other things in device.hpp - maybe worth a follow up once this is in.

PR #164, opened 2026-07-16
Make frameset simple synchronized_value

seanavery:
## Description

There has been an existing race vulnerability between frame callback writer and readers
(get_images, get_point_cloud, and the watchdog loop)

The steady watchdog poll at 1hz surfaced:
```
[module/watchdog.hpp:114] [watchdog] poll error: out of range value for argument "index"
```

latest_frameset_ was a std::shared_ptr<synchronized_value<rs2::frameset>> 

the inner synchronized_value was not working as intended - the whole object was being
replaced via the frame callback, so a reader locked the old mutex while the writer installed a new one

Furthermore, the shared_ptr use seems like it was unnecessary -  rs2::frameset is already a refcounted (does not perform any expensive data copy operations). See class [here](https://github.com/realsenseai/librealsense/blob/fec2d156e531f417c927262818f3440cfbcde4e9/include/librealsense2/hpp/rs_frame.hpp#L1066) and ref increment [here](https://github.com/IntelRealSense/librealsense/blob/fec2d156e531f417c927262818f3440cfbcde4e9/include/librealsense2/hpp/rs_frame.hpp#L411-L419).

## Tests
- [x] long running manual test on amd64 machine to sanity check
- [x] try out on the problematic machine

## Corollary
- [x] figure out why we were not seeing this issue surfaced through standard get_images use

It likely did, invisibly:

librealsense recycles composites from a pool of identically-shaped (color+depth) frames, so a stale handle usually still "works", callers just silently got a recycled frameset. On the rare shape mismatch it threw, `get_images` returned a one-off RPC error that clients retried past.

### DIFF
--- a/src/module/device.hpp
+++ b/src/module/device.hpp
@@ -90,10 +90,9 @@ void deviceChangedCallback(
     std::uint64_t maxFrameAgeMs);
 
 template <typename FrameT, typename FrameSetT, typename ViamConfigT>
-void frameCallback(
-    FrameT const &frame, std::uint64_t const maxFrameAgeMs,
-    boost::synchronized_value<std::shared_ptr<FrameSetT>> &frame_set_,
-    ViamConfigT const &viamConfig);
+void frameCallback(FrameT const &frame, std::uint64_t const maxFrameAgeMs,
+                   boost::synchronized_value<FrameSetT> &frame_set_,
+                   ViamConfigT const &viamConfig);
 
 /********************** DEVICE LIFECYCLE ************************/
 template <typename ViamConfigT, typename ViamDeviceT = ViamRSDevice<>,
@@ -112,11 +111,11 @@ bool destroyDevice(
 
 /********************** STREAMING LIFECYCLE ************************/
 template <typename ViamDeviceT, typename FrameSetT, typename ViamConfigT>
-void startDevice(
-    std::string const &serialNumber,
-    std::shared_ptr<boost::synchronized_value<ViamDeviceT>> dev,
-    std::shared_ptr<boost::synchronized_value<FrameSetT>> &frame_set_storage,
-    std::uint64_t const maxFrameAgeMs, ViamConfigT const &viamConfig);
+void startDevice(std::string const &serialNumber,
+                 std::shared_ptr<boost::synchronized_value<ViamDeviceT>> dev,
+                 boost::synchronized_value<FrameSetT> &frame_set_storage,
+                 std::uint64_t const maxFrameAgeMs,
+                 ViamConfigT const &viamConfig);
 
 template <typename ViamDeviceT>
 bool stopDevice(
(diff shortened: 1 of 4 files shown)


seanavery:
Repro script TSan result:

old pattern = 5–11 races + crash/deadlock in <2 s
 
new = clean
```cpp
// ThreadSanitizer repro for the latest_frameset_ publish/read race.
//
//   old: shared_ptr<synchronized_value<FS>> swapped per frame (the bug)
//   new: one synchronized_value<FS>, value assigned under its lock (the fix)
//
// Mirrors the real code shapes:
//   writer = frameCallback (~30Hz in prod; full-speed here)
//   readers = get_images / get_point_cloud / watchdog get_fs
//
// Expected: `./repro old` -> TSan "data race" reports (exit 66)
//           `./repro new` -> clean (exit 0)
#include <boost/thread/synchronized_value.hpp>

#include <atomic>
#include <chrono>
#include <cstdio>
#include <cstring>
#include <memory>
#include <mutex>
#include <thread>
#include <vector>

// Stand-in for rs2::frameset with a torn-read canary: a and b must match.
struct FS {
  long a = 0;
  long b = 0;
};

using SV = boost::synchronized_value<FS, std::mutex>;

static std::atomic<bool> stop{false};
static std::atomic<long> torn{0};

template <typename WriteFn, typename ReadFn>
int run(WriteFn write, ReadFn read) {
  std::thread writer([&] {
    long n = 0;
    while (!stop.load(std::memory_order_relaxed)) {
      write(FS{n, n});
      ++n;
    }
  });
  std::vector<std::thread> readers;
  for (int i = 0; i < 4; ++i) {
    readers.emplace_back([&] {
      while (!stop.load(std::memory_order_relaxed)) {
        FS f = read();
        if (f.a != f.b)
          torn.fetch_add(1, std::memory_order_relaxed);
      }
    });
  }
  std::this_thread::sleep_for(std::chrono::seconds(2));
  stop.store(true);
  writer.join();
  for (auto &r : readers)
    r.join();
  std::printf("torn reads observed: %ld\n", torn.load());
  return 0;
}

int main(int argc, char **argv) {
  if (argc < 2 || (std::strcmp(argv[1], "old") && std::strcmp(argv[1], "new"))) {
    std::fprintf(stderr, "usage: %s old|new\n", argv[0]);
    return 2;
  }

  if (!std::strcmp(argv[1], "old")) {
    // THE BUG: pointer swapped per frame; readers deref unsynchronized.
    std::shared_ptr<SV> storage;
    return run(
        [&](FS f) { storage = std::make_shared<SV>(f); },       // frameCallback
        [&]() -> FS { return storage ? storage->get() : FS{}; } // readers
    );
  }

  // THE FIX: one box forever; every access goes through its mutex.
  SV storage;
  return run([&](FS f) { storage = f; },        // frameCallback
             [&]() -> FS { return storage.get(); } // readers
  );
}
```

njooma:
I'm going to have to defer to @lia-viam's expertise here

lia-viam (approved):

PR #165, opened 2026-07-20
Add D405 camera support

paigetrexel-vm:
The D405 differs from other supported D400 cameras in two ways that the module previously could not handle:

- It has no dedicated RGB sensor: the single stereo sensor serves both the color and depth streams and does not cast to rs2::color_sensor. Sensor discovery in createSingleSensorConfig/createSwD2CAlignConfig now selects streams by profile format across all sensors instead of looking for dedicated color and depth sensors, and auto-exposure priority is disabled on whichever sensor serves color. Non-video profiles (e.g. IMU motion streams) are skipped explicitly.

- Its default depth units are 0.1 mm rather than 1 mm. GetImage depth encoding now rescales raw depth values to the millimeters that image/vnd.viam.dep encodes, using the frame's reported depth units (pass-through for 1 mm devices). GetPointCloud was already correct (rs2::pointcloud applies the scale internally).

Also adds the D405 entry to get_geometries (42x42x23 mm, left-imager origin, from datasheet 337029-017 Tables 3-54/4-16/4-20; color->depth baseline is ~0 because color comes from the left imager) and documents D405 frame semantics in the README.

Firmware auto-detect needs no change: since the 5.13 unification the D400-series firmware line covers the D405, and librealsense 2.57.7 recommends 5.17.0.10 for all D400 PIDs.

Validated on a physical D405.

### DIFF
--- a/src/module/device_impl.hpp
+++ b/src/module/device_impl.hpp
@@ -10,6 +10,7 @@
 #include <optional>
 #include <string>
 #include <string_view>
+#include <type_traits>
 #include <utility>
 
 #include <viam/sdk/log/logging.hpp>
@@ -141,40 +142,30 @@ void applyDepthSensorOptions(SensorT &sensor, ViamConfigT const &viamConfig,
   }
 }
 
+// Call this on the sensor that serves the color stream. On most D400
+// cameras that is the dedicated RGB sensor; on the D405 the stereo sensor
+// serves color itself, so callers gate on what the sensor serves rather
+// than on its extension type.
 template <typename SensorT>
 void disableAutoExposurePriority(SensorT &sensor,
                                  viam::sdk::LogSource &logger) {
-  try {
-    auto sensor_type = sensors::get_sensor_type(sensor, logger);
-    if (sensor_type == sensors::SensorType::unknown) {
-      throw std::runtime_error("Unknown sensor type");
+  // CRITICAL: Disable auto-exposure priority to maintain frame sync.
+  // When enabled, RGB sensor drops frames to adjust exposure, breaking
+  // temporal alignment with depth (causes 5-20ms timestamp drift).
+  // Tradeoff: slightly worse exposure in changing light conditions,
+  // but tight temporal sync (mostly <5ms) between color and depth.
+  if (sensor.supports(RS2_OPTION_AUTO_EXPOSURE_PRIORITY)) {
+    try {
+      // Disable auto-exposure priority to ensure constant FPS
+      sensor.set_option(RS2_OPTION_AUTO_EXPOSURE_PRIORITY, 0.0);
+      VIAM_DEVICE_LOG(logger, info)
+          << "[disableAutoExposurePriority] Disabled Auto-Exposure Priority "
+             "(constant FPS) for color sensor";
+    } catch (const std::exception &e) {
+      VIAM_DEVICE_LOG(logger, warn) << "[disableAutoExposurePriority] Failed "
+                                       "to disable Auto-Exposure Priority: "
+                                    << e.what();
     }
-    // Only disable auto-exposure priority for color sensors
-    if (sensor_type != sensors::SensorType::color) {
-      return;
-    }
-    // CRITICAL: Disable auto-exposure priority to maintain frame sync.
-    // When enabled, RGB sensor drops frames to adjust exposure, breaking
-    // temporal alignment with depth (causes 5-20ms timestamp drift).
-    // Tradeoff: slightly worse exposure in changing light conditions,
-    // but tight temporal sync (mostly <5ms) between color and depth.
-    if (sensor.supports(RS2_OPTION_AUTO_EXPOSURE_PRIORITY)) {
-      try {
-        // Disable auto-exposure priority to ensure constant FPS
-        sensor.set_option(RS2_OPTION_AUTO_EXPOSURE_PRIORITY, 0.0);
-        VIAM_DEVICE_LOG(logger, info)
-            << "[disableAutoExposurePriority] Disabled Auto-Exposure Priority "
-               "(constant FPS) for color sensor";
-      } catch (const std::exception &e) {
-        VIAM_DEVICE_LOG(logger, warn) << "[disableAutoExposurePriority] Failed "
-                                         "to disable Auto-Exposure Priority: "
-                                      << e.what();
-      }
-    }
-  } catch (const std::exception &e) {
-    VIAM_DEVICE_LOG(logger, error)
-        << "[disableAutoExposurePriority] Failed to get sensor type: "
-        << e.what();
   }
 }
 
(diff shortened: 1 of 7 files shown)


shannonbradshaw (commented):
Reviewed this against #160, which it supersedes: commit 29474c6 is the #160 commit rebased onto current main (authorship preserved), with the conflicts from #147/#159/#161/#162 resolved. A patch-to-patch diff of the two commits shows the README, encoding.{hpp,cpp}, realsense.hpp, and both test files are byte-identical except for line offsets, so the review below focuses on the rebase adaptations, which are the only new code.

**Rebase adaptations (both correct)**

- `createSingleSensorConfig`: the pre-rebase code applied depth-sensor options when the sensor cast to `rs2::depth_sensor`; this now gates on `SensorTypeTraits<SensorT>::stream_type == RS2_STREAM_DEPTH`. That is the right translation for format-based discovery: a depth-only config on a D405 still gets its depth/IR options applied to the stereo sensor, and a color-only config on a D435 correctly skips them.
- `createSwD2CAlignConfig`: adds a `serves_depth` flag alongside `serves_color` and applies `applyDepthSensorOptions` to whichever sensor emits Z16 profiles. On the D405 that is the same sensor that gets auto-exposure priority disabled, which is correct since there is only one sensor.

**Issues to resolve before merge (both minor)**

1. Validation claim inconsistency. The PR body says "Validated on a physical D405," but the commit message still says "Not yet validated on a physical D405," and the README still says "D405 support is implemented ... but has not yet been validated on the platforms above," with an empty D405 column in the hardware table. If the validation was on one of the listed platforms, the table should get an X and the caveat should be dropped or narrowed; if it was on a different platform, the caveat should say what was validated. As written, the docs contradict the PR.
2. Duplicate PRs. #160 is still open with green CI. I'll close it in favor of this one since this branch is rebased onto current main and hardware-validated.

**Notes**

- Both linux test jobs (amd64 and arm64) pass; the macOS jobs were still pending when I reviewed. Nothing in the diff is platform-sensitive beyond what already passed.
- The pre-existing operator-precedence quirk in the resolution match (`(not width) or (width == ...) and (...)`) is carried over unchanged. Not a regression, just a latent oddity if someone configures height without width.

paigetrexel-vm:
> Reviewed this against #160, which it supersedes: commit [29474c6](https://github.com/viam-modules/viam-camera-realsense/commit/29474c62b17f1880f9251d0ee65757de58da2b7a) is the #160 commit rebased onto current main (authorship preserved), with the conflicts from #147/#159/#161/#162 resolved. A patch-to-patch diff of the two commits shows the README, encoding.{hpp,cpp}, realsense.hpp, and both test files are byte-identical except for line offsets, so the review below focuses on the rebase adaptations, which are the only new code.
> 
> **Rebase adaptations (both correct)**
> 
> * `createSingleSensorConfig`: the pre-rebase code applied depth-sensor options when the sensor cast to `rs2::depth_sensor`; this now gates on `SensorTypeTraits<SensorT>::stream_type == RS2_STREAM_DEPTH`. That is the right translation for format-based discovery: a depth-only config on a D405 still gets its depth/IR options applied to the stereo sensor, and a color-only config on a D435 correctly skips them.
> * `createSwD2CAlignConfig`: adds a `serves_depth` flag alongside `serves_color` and applies `applyDepthSensorOptions` to whichever sensor emits Z16 profiles. On the D405 that is the same sensor that gets auto-exposure priority disabled, which is correct since there is only one sensor.
> 
> **Issues to resolve before merge (both minor)**
> 
> 1. Validation claim inconsistency. The PR body says "Validated on a physical D405," but the commit message still says "Not yet validated on a physical D405," and the README still says "D405 support is implemented ... but has not yet been validated on the platforms above," with an empty D405 column in the hardware table. If the validation was on one of the listed platforms, the table should get an X and the caveat should be dropped or narrowed; if it was on a different platform, the caveat should say what was validated. As written, the docs contradict the PR.
> 2. Duplicate PRs. #160 is still open with green CI. I'll close it in favor of this one since this branch is rebased onto current main and hardware-validated.
> 
> **Notes**
> 
> * Both linux test jobs (amd64 and arm64) pass; the macOS jobs were still pending when I reviewed. Nothing in the diff is platform-sensitive beyond what already passed.
> * The pre-existing operator-precedence quirk in the resolution match (`(not width) or (width == ...) and (...)`) is carried over unchanged. Not a regression, just a latent oddity if someone configures height without width.

"Issues" are both cosmetic since I took over and tested this PR from Shannon. I can confirm that I did test this with actual hardware.

src/module/encoding.cpp
```diff
@@ -126,8 +144,20 @@ encodeDepthFrameToResponse(rs2::depth_frame const &frame) {
     throw std::runtime_error("[encodeDepthFrameToResponse] frame profile is "
                              "not a video stream profile");
   }
+  // image/vnd.viam.dep values are millimeters. Most D400 cameras use 1 mm
+  // depth units so the raw values pass through unchanged, but the D405
+  // defaults to 0.1 mm units and must be rescaled.
+  float mm_per_unit = 1.0f;
+  try {
+    const float meters_per_unit = frame.get_units();
```
seanavery, line 152: Nice! Since we can fetch the units value directly from the frame, seems like it should work across all the other camera types

src/module/device_impl.hpp
```diff
@@ -427,45 +418,48 @@ createSingleSensorConfig(std::shared_ptr<DeviceT> dev,
...
+    auto profiles = s.get_stream_profiles();
+    bool sensor_serves_stream = false;
+    for (auto &cp : profiles) {
+      // Skip non-video profiles (e.g. motion streams on IMU-equipped
+      // models) — casting them to a video profile is invalid.
+      if (not cp.template is<VideoStreamProfileT>()) {
+        continue;
+      }
+      auto csp = cp.template as<VideoStreamProfileT>();
+      if (csp.format() != SensorTypeTraits<SensorT>::format_type) {
```
seanavery, line 435: Should we also check that stream_type is color here?

src/module/device_impl.hpp
```diff
@@ -141,40 +142,30 @@ void applyDepthSensorOptions(SensorT &sensor, ViamConfigT const &viamConfig,
...
-        VIAM_DEVICE_LOG(logger, warn) << "[disableAutoExposurePriority] Failed "
-                                         "to disable Auto-Exposure Priority: "
-                                      << e.what();
-      }
+  // CRITICAL: Disable auto-exposure priority to maintain frame sync.
+  // When enabled, RGB sensor drops frames to adjust exposure, breaking
+  // temporal alignment with depth (causes 5-20ms timestamp drift).
+  // Tradeoff: slightly worse exposure in changing light conditions,
+  // but tight temporal sync (mostly <5ms) between color and depth.
+  if (sensor.supports(RS2_OPTION_AUTO_EXPOSURE_PRIORITY)) {
```
seanavery, line 157: we should move this inside the try catch block right?

seanavery (commented):
First pass, looking good to me!
I don't see any regression or compatibility concerns with this change

Created [rc release](https://github.com/viam-modules/viam-camera-realsense/actions/runs/29842703451) `0.22.5-rc0` we can test with to make sure things are working as expected

seanavery:
@paigetrexel-vm Just following up here, any success testing out with the D405?

PR #168, opened 2026-07-28
fix(watchdog): reattach the camera after it drops off the bus

NickPPC:
## Summary

A RealSense camera that drops off the USB bus *after* the resource is constructed leaves the module alive but permanently dead — recovering it requires rebooting the host. The stale-frame watchdog's only remedy is rebuilding the `rs2::pipeline`, which cannot help when there is no device to rebuild, so it logs `skip restart: recovery mode or no device` forever and nothing ever re-polls `query_devices()`. This PR gives the watchdog a second remedy: detect that the device is gone and re-enumerate the bus until it comes back.

Observed in production on machine part `5be4df6e-b9ae-4a5f-acc0-e258d51be17d` (D435I, serial `261322075626`): the camera dropped at 14:23Z and again at 20:15Z UTC, each time wedging until a reboot. Across the outage the module logged 2103 `sustained stale frame` warnings and 524 `skip restart` lines without ever attempting recovery, and the dead `cam` resource cascaded into every dependent service (`cup-detection-segmenter`, `coffee-lifecycle`, `maintenance-sensor`, `stream-deck`) failing to build.

## Changes

- **src/module/watchdog.hpp**: Add optional `DeviceAbsentFn` / `ReattachFn` callbacks and a `reattach_interval_ms` tunable (5 s default); `maybe_reattach()` retries indefinitely, deliberately outside the 6/hour restart cap, since an unplugged camera can reappear at any time.
- **src/module/watchdog.hpp**: Check device-absence *before* the frame-age gate in `poll_once` — a device that dropped off before it ever streamed has no frame to age, so the stale path would never fire for it — and *after* `recovery_check_`, so a DFU-mode device stays in the firmware-update flow.
- **src/module/watchdog.hpp**: Throttle the "no device" warning to the first failure plus once a minute, replacing the per-poll `sustained stale frame` spam for the entire duration of an outage.
- **src/module/realsense.hpp**: Wire `device_absent` (`!device_`) and `reattach` (`query_devices()` + `assign_and_initialize_device()`, the same call shape the constructor uses) into the watchdog.
- **src/module/realsense.hpp**: Add `attach_mutex_` and take it in `deviceChangedCallback`. Both that callback and the new reattach path call `assign_and_initialize_device`, and for an explicitly configured serial the `assigned_serials_` check does not stop two threads claiming the same camera. Kept separate from `do_command_mutex_` so librealsense's callback thread never blocks behind a multi-minute firmware update; lock order is `do_command_mutex_` → `attach_mutex_`, never reversed.
- **test/watchdog_tests.cpp**: 8 new tests covering the reattach path.

Both new callbacks are optional trailing constructor parameters, so existing call sites are unchanged.

### Secondary fix

`handleFirmwareUpdate` sets `device_ = nullptr` and comments that the device "will be reassigned when detected by `deviceChangedCallback`", but that callback is cleared for the duration of the update. A camera that finishes rebooting inside that window is missed and never reattaches. The new reattach path covers this case too.

## Testing

New tests in `test/watchdog_tests.cpp`:

- `ReattachesWhenDeviceAbsent` — the core regression
- `DoesNotRestartWhileDeviceAbsent` — a restart must not be attempted with no device
- `ReattachesWithNoFramesEverDelivered` — absence detected without a frame to age
- `StopsReattachingOnceDevicePresent` — no re-enumeration storm after recovery
- `SkipsReattachDuringRecoveryMode` — does not fight a firmware update
- `ReattachIsThrottledToItsInterval` — bounds bus-enumeration cost
- `SurvivesReattachException` — a throwing `rs2` enumeration must not kill the thread
- `WorksWithoutReattachCallbacks` — omitting both preserves prior behavior

**Verified:** `watchdog.hpp` — 19/19 tests pass (11 existing + 8 new), clean under `-Wall -Wextra -Wreorder -Werror`, 5 repeat runs green, ThreadSanitizer reports 0 warnings. Mutation-tested: moving the device-absent check below the frame-age gate fails `ReattachesWithNoFramesEverDelivered` and nothing else. `clang-format-19` reports all three files already correctly formatted.

**Not verified — needs CI:** `realsense.hpp` was not compiled. conan was unavailable on the machine used to write this, so the tests above were built against a stub gtest and a stub `viam/sdk/log/logging.hpp` rather than through `make test`. The reattach lambda mirrors the constructor's existing call shape, but a real build is required before merge.

**Suggested manual verification:** with the module streaming, unplug the camera and confirm the log shows one `no device to stream from` warning rather than continuous `sustained stale frame` spam; replug and confirm `device reattached` appears and streaming resumes without restarting `viam-server` or the host.

## Claude Code Prompts Used

- "Hi Claude, we have been having some issues with our camera losing connectivity on the Viam machine part 5be4df6e-b9ae-4a5f-acc0-e258d51be17d. Can you check if there is any way for us to update the camera firmware which is exposed through DoCommand? Can you also check the logs on that machine to see if you can understand why we lost connectivity and why we did not get it back until we rebooting the computer?"
- "Let's draft a patch for #3"
- "continue"
- "open a draft PR"


seanavery (approved):
Looks good to me.

v1 of the watchdog purposefully punted on device recovery; it only handled a wedged pipeline on a still attached device

unplug/re-plug should already be handled by deviceChangedCallback. Wondering exactly how we get into a state where device_ is null and stays null. Does the event never fire, or does it fire and the re-assign fail?

NickPPC:
> Looks good to me.
> 
> v1 of the watchdog purposefully punted on device recovery; it only handled a wedged pipeline on a still attached device
> 
> unplug/re-plug should already be handled by deviceChangedCallback. Wondering exactly how we get into a state where device_ is null and stays null. Does the event never fire, or does it fire and the re-assign fail?

Unclear how we get there but we have had many instances of the camera getting "disconnected" and where unplugging and plugging it back in resolved the issue. The issue usually subside when we change the cable, so it might be a bad cable or a USB protocol downgrade which does not count as being fully disconnected 🤷 

I will hot reload this on Cappuccina for the rest of the week and we should merge if we do not notice any adverse effect

seanavery:
Hey @NickPPC any update on the testing here? Do we have logs showing successful device recovery?

NickPPC:
> Hey Nicolas Palpacuer (@NickPPC) any update on the testing here? Do we have logs showing successful device recovery?

It was running on Cappuccina for a bit but it does not seem to be on anymore. I will hot reload it and let it sit for a week to be sure we do not have a regression and to see if we can catch an instance of the reattachment of the camera

NickPPC:
> Hey Nicolas Palpacuer (@NickPPC) any update on the testing here? Do we have logs showing successful device recovery?

@seanavery I have seen an example of a successful recovery of this tuype yet (maybe we do not have a flaky camera connection on Cappuccina anymore). However there has been no visible downsides so I will be merging this

PR #170, opened 2026-08-06
fix(properties): report identity extrinsics for the color-registered cloud

HipsterBrown:
## What

`get_properties()` now reports **identity** extrinsics instead of the depth→color baseline.

## Why

#161 (shipped in 0.22.4) made `get_point_cloud()` align depth to the color frame. But `get_properties()` kept advertising the depth→color offset, so the driver was telling consumers the cloud was in a frame it no longer sits in.

Anything that applies that offset before projecting — including RDK's `camera.Properties.PointToPixel` — re-corrects for a baseline `rs2::align` already removed. Result: every projected point lands **~22 px off horizontally**, in the wrong direction.

The point cloud is already in the color frame. The intrinsics we report are the color intrinsics. Same frame ⇒ the transform between them is identity.

## Verified on hardware

Live D435 (part `96657d75`), before and after reload:

| | 0.22.4 | This PR |
|---|---|---|
| `translation.x` | −15.044 | **0** |
| `translation.y` | −0.173 | **0** |
| `translation.z` | −0.428 | **0** |
| `focalXPx` | 922.700 | 922.700 (unchanged) |

Intrinsics are byte-identical, confirming only the extrinsics path changed.

`147/147` tests pass (144 on main — the placeholder test file became 4 real tests). Lint clean.

## Does this undo #161?

No. The alignment stays exactly as it was. #161 fixed consumers that project *without* extrinsics (`detections-to-segments`); this fixes the ones that *do*. Identity + a color-registered cloud is the only state that satisfies both.

---

<details>
<summary><b>Why identity is correct for every <code>sensors</code> config (not just the default)</b></summary>

<br>

I checked each combination rather than hardcoding identity:

| `sensors` | cloud emitted in | before | after |
|---|---|---|---|
| `["color","depth"]` | color (`rs2::align`) | color→depth (−15 mm) ❌ | identity |
| `["color"]` | no cloud — `get_point_cloud` throws on the missing depth frame | identity ✅ | identity |
| `["depth"]` | no cloud — `PointCloudFilter` throws on the missing color frame | identity ✅ | identity |

With one sensor there is no second frame to be in, and no point cloud at all. Those cases already returned identity via the existing same-stream path in `get_extrinsics`, so **only the both-sensors case changes behavior**.

</details>

<details>
<summary><b>How three PRs drifted apart</b></summary>

<br>

- **#141** anchored extrinsics to the depth left imager
- **#148** anchored `get_geometries` to the color sensor
- **#161** registered the point cloud to the color frame

Only #141's extrinsics were left behind. The two comments even contradicted each other in the source — `get_properties` claimed "the camera reference frame is always the depth left imager, matching get_geometries", while `get_geometries` claimed the origin is the color sensor.

After this PR the intrinsics, the point cloud, and the geometry box pose all agree on one frame.

</details>

<details>
<summary><b>Is there another extrinsics consumer? (checked — no)</b></summary>

<br>

`get_extrinsics` had **exactly one caller**: the line this PR changes.

The comment block at `realsense.hpp:918-935` derives the `get_geometries` box pose from the `~= {-14.7, 0, 0} mm` color→depth baseline, but that is documentation only — the poses are hardcoded constants and nothing reads extrinsics at runtime. So there was no internal value that needed preserving. I added a clarifying line so it doesn't read as a runtime dependency.

</details>

<details>
<summary><b>Why the tests are shaped this way</b></summary>

<br>

`get_properties()` isn't unit-testable — it needs an active profile off a live `rs2::pipeline`, which is why `extrinsics_test.cpp` was a placeholder.

So `get_extrinsics` is now templated on the profile type (the same injection idiom `device.hpp` already uses) and tests drive it through a stand-in exposing `unique_id()`/`get_extrinsics_to()`.

The important part: one test asserts a real baseline between two *distinct* streams still converts m→mm correctly. Without it, the identity assertions could pass simply because the function always returned zero.

</details>

<details>
<summary><b>Known, not fixed here: ~6 mm calibration under-correction</b></summary>

<br>

The reported extrinsic appears to under-correct by ~6 mm even back when it was the right thing to apply. Pre-#161 captures fit a translation of (−19.5, −3.5) to (−21.5, −1.0) mm against the reported (−15.0, 0).

That is a 2-DOF fit with non-zero residual, so it reads as a calibration error rather than a calibrated value. Flagging it; out of scope for this PR. Expect residual divergence downstream to not quite reach zero.

</details>

### DIFF
--- a/src/module/extrinsics.hpp
+++ b/src/module/extrinsics.hpp
@@ -13,9 +13,13 @@ namespace extrinsics {
 /// @return Extrinsic parameters representing the transformation from
 /// from_stream to to_stream. If streams are identical, returns identity
 /// extrinsics.
+///
+/// Templated on the profile types so tests can substitute a stand-in exposing
+/// unique_id() and get_extrinsics_to(); a real rs2::stream_profile can only be
+/// handed out by a live device.
+template <typename FromStreamT, typename ToStreamT>
 inline viam::sdk::Camera::extrinsic_parameters
-get_extrinsics(const rs2::stream_profile &from_stream,
-               const rs2::stream_profile &to_stream) {
+get_extrinsics(const FromStreamT &from_stream, const ToStreamT &to_stream) {
   viam::sdk::Camera::extrinsic_parameters extrinsics;
 
   // If streams are the same, return identity extrinsics
@@ -45,6 +49,30 @@ get_extrinsics(const rs2::stream_profile &from_stream,
   extrinsics.orientation.theta = 0.0;
 
   return extrinsics;
+}
+
+/// @brief The extrinsic parameters get_properties() publishes.
+/// @param reference_stream The stream whose intrinsics get_properties() reports
+/// @return Identity extrinsics.
+///
+/// extrinsic_parameters describes the transform from the frame the point cloud
+/// is emitted in to the frame the reported intrinsics describe. Those are the
+/// same frame: get_point_cloud() aligns depth to color before deprojecting (see
+/// device::PointCloudFilter), so the cloud lands in the color frame, which is
+/// also the stream get_properties() takes its intrinsics from whenever color is
+/// configured. When only one sensor is configured there is no second frame to
+/// be in — and no point cloud at all, since deprojection needs both streams.
+/// Either way the transform is @p reference_stream to itself: identity.
+///
+/// Do NOT publish the real depth<->color baseline here. Consumers subtract
+/// extrinsic_parameters.translation before projecting a cloud point through the
+/// reported intrinsics (rdk's camera.Properties.PointToPixel), so a non-zero
+/// translation re-applies a baseline rs2::align has already removed, shifting
+/// every projected point by fx * T / Z — about 22 px at 0.63 m on a D435.
+template <typename StreamProfileT>
+inline viam::sdk::Camera::extrinsic_parameters
+get_reported_extrinsics(const StreamProfileT &reference_stream) {
+  return get_extrinsics(reference_stream, reference_stream);
 }
 
 } // namespace extrinsics
--- a/src/module/realsense.hpp
+++ b/src/module/realsense.hpp
@@ -787,8 +787,7 @@ public:
       }
       auto fillResp = [this](viam::sdk::Camera::properties &p,
                              rs2_intrinsics const &props,
-                             const rs2::stream_profile &stream,
-                             const rs2::stream_profile &ref_stream) {
+                             const rs2::stream_profile &reference_stream) {
         p.supports_pcd = true;
         p.intrinsic_parameters.width_px = props.width;
         p.intrinsic_parameters.height_px = props.height;
@@ -797,9 +796,10 @@ public:
         p.intrinsic_parameters.center_x_px = props.ppx;
         p.intrinsic_parameters.center_y_px = props.ppy;
 
-        // Calculate extrinsics from stream to reference stream
+        // Identity: the point cloud is emitted in the same frame these
+        // intrinsics describe. See extrinsics::get_reported_extrinsics.
         p.extrinsic_parameters =
-            realsense::extrinsics::get_extrinsics(stream, ref_stream);
+            realsense::extrinsics::get_reported_extrinsics(reference_stream);
 
         switch (props.model) {
         case RS2_DISTORTION_BROWN_CONRADY:
@@ -878,21 +878,23 @@ public:
               << e.what();
         }
 
-        // The camera reference frame is always the depth left imager, matching
-        // get_geometries. Intrinsics come from color when configured (the
-        // stream most callers derive poses from), otherwise depth. Extrinsics
-        // are therefore intrinsics_stream -> depth, which is identity whenever
-        // depth is the intrinsics stream or only one sensor is configured.
-        const rs2::video_stream_profile &intrinsics_stream =
+        // The camera reference frame is the color imager when color is
+        // configured (the stream most callers derive poses from), otherwise
+        // depth. Everything this resource reports is expressed in that one
+        // frame: the intrinsics below, the point cloud get_point_cloud() emits
+        // (it aligns depth to color before deprojecting — see
+        // device::PointCloudFilter), and the box pose from get_geometries().
+        // Extrinsics are therefore identity; see
+        // extrinsics::get_reported_extrinsics for why publishing the real
+        // depth<->color baseline here double-corrects consumers.
+        const rs2::video_stream_profile &reference_stream =
             color_stream ? color_stream : depth_stream;
-        if (not intrinsics_stream) {
+        if (not reference_stream) {
           throw std::runtime_error(
               "neither color nor depth stream is available");
         }
-        const rs2::video_stream_profile &ref_stream =
-            depth_stream ? depth_stream : color_stream;
-        auto props = intrinsics_stream.get_intrinsics();
-        fillResp(response, props, intrinsics_stream, ref_stream);
+        auto props = reference_stream.get_intrinsics();
+        fillResp(response, props, reference_stream);
       } // End scope for my_dev lock
 
       VIAM_RESOURCE_LOG(debug) << "[get_properties] end";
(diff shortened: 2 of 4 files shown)


src/module/realsense.hpp
```diff
@@ -797,9 +796,10 @@ class Realsense final : public viam::sdk::Camera {
         p.intrinsic_parameters.center_x_px = props.ppx;
         p.intrinsic_parameters.center_y_px = props.ppy;
 
-        // Calculate extrinsics from stream to reference stream
+        // Identity: the point cloud is emitted in the same frame these
+        // intrinsics describe. See extrinsics::get_reported_extrinsics.
         p.extrinsic_parameters =
-            realsense::extrinsics::get_extrinsics(stream, ref_stream);
+            realsense::extrinsics::get_reported_extrinsics(reference_stream);
```
HipsterBrown, line 802: **This is the actual fix — one call swapped.**

Was `get_extrinsics(stream, ref_stream)`, where `ref_stream` was the depth stream. That published the ~15 mm depth→color baseline.

Since #161 the cloud is already aligned to color, so that baseline gets applied twice by anything using `PointToPixel`.

HipsterBrown (commented):
Inline notes on the specific decisions, so the description can stay short.

src/module/realsense.hpp
```diff
@@ -878,21 +878,23 @@ class Realsense final : public viam::sdk::Camera {
...
+        // The camera reference frame is the color imager when color is
+        // configured (the stream most callers derive poses from), otherwise
+        // depth. Everything this resource reports is expressed in that one
+        // frame: the intrinsics below, the point cloud get_point_cloud() emits
+        // (it aligns depth to color before deprojecting — see
+        // device::PointCloudFilter), and the box pose from get_geometries().
+        // Extrinsics are therefore identity; see
+        // extrinsics::get_reported_extrinsics for why publishing the real
+        // depth<->color baseline here double-corrects consumers.
+        const rs2::video_stream_profile &reference_stream =
```
HipsterBrown, line 890: Note what's **removed** here: the old code had a second variable,

```cpp
const rs2::video_stream_profile &ref_stream =
    depth_stream ? depth_stream : color_stream;
```

It's gone, and `intrinsics_stream` is renamed `reference_stream` because it's now the single frame everything is reported in.

Deliberate: with no depth stream in scope, there's no wrong value left to accidentally pass to the extrinsics call.

src/module/realsense.hpp
```diff
@@ -878,21 +878,23 @@ class Realsense final : public viam::sdk::Camera {
               << e.what();
         }
 
-        // The camera reference frame is always the depth left imager, matching
-        // get_geometries. Intrinsics come from color when configured (the
-        // stream most callers derive poses from), otherwise depth. Extrinsics
-        // are therefore intrinsics_stream -> depth, which is identity whenever
-        // depth is the intrinsics stream or only one sensor is configured.
-        const rs2::video_stream_profile &intrinsics_stream =
+        // The camera reference frame is the color imager when color is
```
HipsterBrown, line 881: The old comment here claimed *"the camera reference frame is always the depth left imager, **matching get_geometries**"* — which was flatly contradicted by `get_geometries` itself (line ~914), where the comment says the origin is the color sensor.

That contradiction is the bug in miniature. Rewritten to state the one frame everything actually uses.

src/module/realsense.hpp
```diff
@@ -915,7 +917,11 @@ class Realsense final : public viam::sdk::Camera {
     // intrinsics and the images/derived poses callers work with are in the
     // color frame. Box dimensions and the depth-imager placement come from the
     // Intel RealSense D400-Series Datasheet (337029-005); the color sensor's
-    // position is then derived from the color->depth extrinsics.
+    // position is then derived from the color->depth extrinsics. Those are the
+    // physical baseline librealsense reports for the model, used here only to
```
HipsterBrown, line 921: Clarification only — no behavior change.

This block cites the color→depth baseline to justify the hardcoded box pose, which made it look like `get_geometries` might depend on what `get_properties` publishes. It doesn't: the poses below are literal constants and nothing here reads extrinsics at runtime.

I verified this before changing the extrinsics — `get_extrinsics` had exactly one caller.

src/module/extrinsics.hpp
```diff
@@ -47,5 +51,29 @@ get_extrinsics(const rs2::stream_profile &from_stream,
...
+///
+/// Do NOT publish the real depth<->color baseline here. Consumers subtract
+/// extrinsic_parameters.translation before projecting a cloud point through the
+/// reported intrinsics (rdk's camera.Properties.PointToPixel), so a non-zero
+/// translation re-applies a baseline rs2::align has already removed, shifting
+/// every projected point by fx * T / Z — about 22 px at 0.63 m on a D435.
+template <typename StreamProfileT>
+inline viam::sdk::Camera::extrinsic_parameters
+get_reported_extrinsics(const StreamProfileT &reference_stream) {
+  return get_extrinsics(reference_stream, reference_stream);
```
HipsterBrown, line 75: Passing the stream to **itself** is intentional, not a typo.

It routes into the existing same-stream branch above (line ~25), which already returns identity. Reusing that path rather than writing a second identity means there's only one definition of "identity extrinsics" in the codebase.

The wrapper exists to give that intent a name and a place to hang the explanation.

src/module/extrinsics.hpp
```diff
@@ -13,9 +13,13 @@ namespace extrinsics {
 /// @return Extrinsic parameters representing the transformation from
 /// from_stream to to_stream. If streams are identical, returns identity
 /// extrinsics.
+///
+/// Templated on the profile types so tests can substitute a stand-in exposing
+/// unique_id() and get_extrinsics_to(); a real rs2::stream_profile can only be
+/// handed out by a live device.
+template <typename FromStreamT, typename ToStreamT>
```
HipsterBrown, line 20: Templating is purely to make this testable — no behavior change, and `rs2::stream_profile` still deduces fine.

`get_properties()` can't be unit-tested (it needs an active profile off a live `rs2::pipeline`), which is why `extrinsics_test.cpp` was a placeholder before this PR. Templating on the profile type lets tests pass a stand-in exposing `unique_id()`/`get_extrinsics_to()`.

Same injection idiom `device.hpp` already uses for `DeviceT`/`PipeT`.

test/extrinsics_test.cpp
```diff
@@ -4,10 +4,102 @@
...
+  EXPECT_DOUBLE_EQ(extrinsics.translation.x(), 0.0);
+  EXPECT_DOUBLE_EQ(extrinsics.translation.y(), 0.0);
+  EXPECT_DOUBLE_EQ(extrinsics.translation.z(), 0.0);
+}
+
+} // namespace
+
+// Guards the regression tests below against being vacuously true: a non-zero
+// baseline between two *different* streams must still come through, in mm.
+TEST(ExtrinsicsTest, ConvertsMetersToMillimetersBetweenDistinctStreams) {
```
HipsterBrown, line 54: **Please don't drop this test** — it's what keeps the other three honest.

The identity assertions below would pass just as happily if `get_extrinsics` were broken and always returned zero. This one proves a real baseline between two *distinct* streams still comes through, correctly converted m→mm.

The values are the actual measurements from the affected D435 (part `96657d75`): `-15.044, -0.173, -0.428`.

test/extrinsics_test.cpp
```diff
@@ -4,10 +4,102 @@
...
+// get_point_cloud() aligns depth to the color frame before deprojecting, so the
+// cloud is already registered to the stream whose intrinsics get_properties()
+// reports. Publishing the depth<->color baseline made consumers that subtract
+// extrinsic_parameters.translation before projecting (rdk's
+// camera.Properties.PointToPixel) re-apply a baseline rs2::align had already
+// removed — a ~22 px X shift at 0.63 m on a D435.
+//
+// The reference stream is color when color is configured, depth otherwise; in
+// every case the reported extrinsics must be identity.
+TEST(ExtrinsicsTest, ReportedExtrinsicsAreIdentityForColorAndDepthConfigured) {
```
HipsterBrown, line 86: The regression test for this PR.

The fake is deliberately booby-trapped: its `get_extrinsics_to()` returns the real −15 mm D435 baseline. If someone reverts to publishing that baseline, this fails.

Pre-fix, the equivalent call returned −15.044; now it must be 0.

seanavery (approved):
Thanks for the detailed follow-up! Makes sense to return identity extrinsics now that we align the pointcloud to color

The API for properties and pointcloud seems pretty akward to me.. When you call get_properties you just assume you are getting the intrinsics/extrinsics for whatever frame the pointcloud happens to be in. Nothing tells you which sensor that is or lets you ask for a different one

